PR #5, opened 2016-06-13
skipping non parseable values

Phantuman:
version and libevent would not show up in Graphite.
libevent would throw an error in grafsy

### DIFF
--- a/memcached.py
+++ b/memcached.py
@@ -46,7 +46,12 @@ def main(argv=None):
     template = 'servers.' + hostname + '.software.memcached.{1} {2} ' + timestamp
 
     for key in memcached:
-      print(template.format(hostname,key, memcached[key]))
+      try:
+        float(memcached[key])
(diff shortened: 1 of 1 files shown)


memcached.py
```diff
@@ -46,7 +46,12 @@ def main(argv=None):
     template = 'servers.' + hostname + '.software.memcached.{1} {2} ' + timestamp
 
     for key in memcached:
-      print(template.format(hostname,key, memcached[key]))
+      try:
+        float(memcached[key])
+        print(template.format(hostname,key, memcached[key]))
+      except ValueError:
+        pass
+
```
hasegeli: I don't think `try: except:` is appropriate in here.  How about changing the regex up in the file to match only with decimal numbers?

This script was so badly written, I couldn't stop myself from reworking it.  Could you please look at my pull request instead?

https://github.com/innogames/igcollect/pull/6/commits

PR #7, opened 2016-07-18
Improving the fastly script based on issue TECH-8655

TheFRedFox:
Greatly increases the performance of the script by highly reducing the amount of requests to get all the information.

Also made it more reliable by increasing the timeout.

[TECH-8655](https://jira.innogames.de/browse/TECH-8655)

### DIFF
--- a/fastly.py
+++ b/fastly.py
@@ -55,8 +55,8 @@ def main(args):
     # Just show a list of possible services
     if args.show_list:
         all_services = get_services()
-        for service in all_services:
-            print(service)
+        for service_id, service_name in all_services.items():
+            print('{}:{}'.format(service_name, service_id))
         sys.exit(0)
 
     # Query the API for all regions and print the list of them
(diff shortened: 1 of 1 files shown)


fastly.py
```diff
@@ -11,119 +11,204 @@
...
 
-    # use the provided start and end if present
-    if args.end_time: endtime = args.end_time
-    if args.start_time: starttime = args.start_time
+                for entry in data:
+                    for key in entry:
+                        value = format_key(entry, key)
+                        if not value:
+                            continue
+                        print(output.format(value=value))
```
hasegeli: How about
```python
if value:
    print(output.format(value=value))
```

TheFRedFox: I prefer the "interesting" part in the "main" indention level. If, for whatever reasons, someone will add more functionality besides the `print`, then it is easier to read in my opinion. :wink:

PR #19, opened 2016-11-22
add virtual memory page stats script

juergenthomann:
keeps track of:
- pages in
- pages out
- minor page faults
- major page faults

### DIFF
--- a/linux_vmstat.py
+++ b/linux_vmstat.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+#
+# igcollect - Linux virtual memory stats
+#
+# Copyright (c) 2016, InnoGames GmbH
+#
+
+import time
+from socket import gethostname
+from argparse import ArgumentParser
+
+
+def parse_args():
+    parser = ArgumentParser(prog='linux_vmstat.py')
+    parser.add_argument(
+        '--fields',
+        dest='fields',
+        default=[],
+        nargs='*',
+        help='The fields from /proc/vmstat to send',
+    )
+    return vars(parser.parse_args())
+
+
+# utility to read and parse a space delimited file (meminfo)
+def parse_split_file(filename):
+    with open(filename, 'r') as f:
+        return [line.strip().split(None, 1) for line in f]
+
+
+def get_vmstat():
+    lines = parse_split_file('/proc/vmstat')
+    return {key: int(value.split()[0]) for key, value in lines}
+
+
+def main(fields):
+    vmstat = get_vmstat()
+    timestamp = int(time.time())
+    hostname = gethostname().replace('.', '_')
+
+    template = "servers.{0}.system.virtualmemory.{1} {2} {3}"
+    for field in fields:
+        print(template.format(hostname, field, vmstat[field], timestamp))
+
+if __name__ == '__main__':
+    main(**parse_args())


linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+#
+# igcollect - Linux virtual memory stats
+#
+# Copyright (c) 2016, InnoGames GmbH
+#
+
+from __future__ import print_function
```
hasegeli: This is not necessary.

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+from socket import gethostname
+import time, sys
```
hasegeli: PEP 8 says they should go to separate lines.

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+
+# utility to read and parse a space delimited file (meminfo)
+def parse_split_file(filename):
+    try:
+        with open(filename, 'r') as f:
+            return [line.strip().split(None, 1) for line in f]
+    except:
```
hasegeli: It is always a bad idea to `expect` anything.  You would hide things like syntax errors, or Ctrl+C, if they happen to be inside this.  I don't thing we need any reason to expect anything in here.

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+vmstat = get_vmstat()
+timestamp = str(int(time.time()))
+hostname = gethostname().replace('.', '_')
+
+# Define desired fileds for ouput
+used_fields = ['pgfault', 'pgmajfault', 'pgpgin', 'pgpgout']
+
+template = "servers.{0}.system.virtualmemory.{1} {2} {3}"
+for field in used_fields:
+    print(template.format(hostname, field, vmstat[field], timestamp))
```
hasegeli: It is better style to put the rest inside `main()`.

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+def get_vmstat():
+    lines = parse_split_file('/proc/vmstat')
+    return dict((key, ( int(value.split()[0]))) for key, value in lines)
+
+vmstat = get_vmstat()
+timestamp = str(int(time.time()))
+hostname = gethostname().replace('.', '_')
+
+# Define desired fileds for ouput
+used_fields = ['pgfault', 'pgmajfault', 'pgpgin', 'pgpgout']
```
hasegeli: I would make this an argument.

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+def parse_split_file(filename):
+    try:
+        with open(filename, 'r') as f:
+            return [line.strip().split(None, 1) for line in f]
+    except:
+        sys.exit(1)
+
+def get_vmstat():
+    lines = parse_split_file('/proc/vmstat')
+    return dict((key, ( int(value.split()[0]))) for key, value in lines)
```
hasegeli: You can also use dict comprehension syntax: `{key: int(value.split()[0]) for key, value in lines}`

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+import time, sys
+
+# utility to read and parse a space delimited file (meminfo)
+def parse_split_file(filename):
+    try:
+        with open(filename, 'r') as f:
+            return [line.strip().split(None, 1) for line in f]
+    except:
+        sys.exit(1)
+
```
hasegeli: PEP 8 says two new lines between functions.

linux_vmstat.py
TheFRedFox: 22:24: E201 whitespace after '('

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# Copyright (c) 2016, InnoGames GmbH
+#
+
+from __future__ import print_function
+from socket import gethostname
+import time, sys
+
+# utility to read and parse a space delimited file (meminfo)
+def parse_split_file(filename):
```
TheFRedFox: PEP8: 13:1: E302 expected 2 blank lines, found 1

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+            return [line.strip().split(None, 1) for line in f]
+    except:
+        sys.exit(1)
+
+def get_vmstat():
+    lines = parse_split_file('/proc/vmstat')
+    return dict((key, ( int(value.split()[0]))) for key, value in lines)
+
+vmstat = get_vmstat()
+timestamp = str(int(time.time()))
```
TheFRedFox: IMO timestamp should be an `int` and not a `string`. You should safely use `int` below in the `format` methods.

linux_vmstat.py
```diff
@@ -0,0 +1,33 @@
...
+hostname = gethostname().replace('.', '_')
+
+# Define desired fileds for ouput
```
TheFRedFox: type "fileds" => "files" and "ouput" => "output"

linux_vmstat.py
TheFRedFox: What about when one has no read access for the file. Will be raise an error then? Maybe we should just let it be raised.

hasegeli: Sure.  Python raise exceptions for all kinds of failures.

PR #21, opened 2016-12-01
add calculation of freeable space by optimize

elmatze79:


mysql.py
```diff
@@ -25,4 +25,13 @@
...
     if row[1].isdigit():
         print "servers.{0}.software.mysql.variables.{1} {2} {3}".format(hostname, row[0], row[1], now)
-
+# Find out how much space we can recover by Optimize
+sysdbs=['information_schema', 'performance_schema', 'mysql', 'sys', 'test']
+free = 0
+cur.execute("SHOW DATABASES")
+for row in cur.fetchall():
+    if row[0] not in sysdbs:
+        cur.execute('select round(DATA_FREE/1024/1024) from information_schema.tables where TABLE_SCHEMA=\'{0}\' and DATA_FREE>0'.format(row[0]))
```
TheFRedFox: That is not safe for sql injections. Rather use this:
`cur.execute('select ... where TABLE_SCHEMA=%s and ...', [row[0]])

mysql.py
```diff
@@ -25,4 +25,13 @@
...
+        for value in cur.fetchall():
+            free += value[0]
+print "servers.{0}.software.mysql.status.optimize_freeable {1} {2}".format(hostname, free, now)
```
TheFRedFox: @hasegeli You have to note that for you `prefix` changes.

mysql.py
```diff
@@ -25,4 +25,13 @@
 for row in cur.fetchall():
     if row[1].isdigit():
         print "servers.{0}.software.mysql.variables.{1} {2} {3}".format(hostname, row[0], row[1], now)
-
+# Find out how much space we can recover by Optimize
+sysdbs=['information_schema', 'performance_schema', 'mysql', 'sys', 'test']
+free = 0
+cur.execute("SHOW DATABASES")
+for row in cur.fetchall():
+    if row[0] not in sysdbs:
```
TheFRedFox: Rather use `continue` check:
``` python
if row[0] in sysdbs:
    continue
```

elmatze79: Why? What benefit does it have compared to the original?

mysql.py
TheFRedFox: Yeah, it might be really hard to have a database with a name which will execute anything. Quite impossible. However in my opinion we should stick to correct parametrization to be consistent to the secure way. 😉

PR #25, opened 2016-12-16
Add a script for switch port traffic

TuxPowered42:
This is a reincarnation of my rather old script. Body is old and ugly, but I made it look like a modern igcollect script.

### DIFF
--- a/src/switch_snmp.py
+++ b/src/switch_snmp.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python
+#
+# igcollect - Switch port traffic over SNMP
+#
+# Copyright (c) 2016, InnoGames GmbH
+#
+
+from __future__ import print_function
+from argparse import ArgumentParser
+
+import re
+import subprocess
+import time
+
+port_name_oid = '.1.3.6.1.2.1.31.1.1.1.1'
+procurve_trunk_oid = '.1.3.6.1.4.1.11.2.14.11.5.1.7.1.3.1.1.8'
+cisco_agg_oid = '.1.2.840.10006.300.43.1.2.1.1.12'
+
+COUNTERS = {
+    'bytesIn': '.1.3.6.1.2.1.31.1.1.1.6',
+    'bytesOut': '.1.3.6.1.2.1.31.1.1.1.10',
+    'pktsIn': '.1.3.6.1.2.1.31.1.1.1.7',
+    'pktsOut': '.1.3.6.1.2.1.31.1.1.1.11',
+    'brdPktsIn': '.1.3.6.1.2.1.31.1.1.1.9',
+    'brdPktsOut': '.1.3.6.1.2.1.31.1.1.1.13',
+    'ifInErrors': '.1.3.6.1.2.1.2.2.1.14',
+    'ifOutErrors': '.1.3.6.1.2.1.2.2.1.20',
+    'ifInDiscards': '.1.3.6.1.2.1.2.2.1.13',
+    'ifOutDiscards': '.1.3.6.1.2.1.2.2.1.19',
+}
+
+
+def get_snmp_value(host, community, node_name):
+    """ Function retriving value of given command.
+        In the current version only one row response is handled:
+        return last value retireved via SNMP
+    """
+
+    proc = subprocess.Popen(
+        ['snmpwalk', '-On', '-c', community, '-v2c', host, node_name],
+        stdout=subprocess.PIPE,
+    )
+
+    resp = proc.stdout.read()
+    rex = re.compile(r'^([0-9\.]+) = ((.*): )?"?([^"]*)')
+    rem = rex.match(resp)
+
+    if rem:
+        return rem.groups()[3]
+    return None
+
+
+def get_snmp_table(host, community, table_oid, skip):
+    ''' Retrieves whole table of SNMP values at once.
+        This should be faster than snmpwalk/get on each OID separately.
+        If a non-null skip is given, last number in OID is compared
+        to the parameter and skipped if bigger.
+    '''
+
+    snmp_proc = subprocess.Popen(
+        ['snmpwalk', '-On', '-c', community, '-v2c', host, table_oid],
+        stdout=subprocess.PIPE,
+    )
+
+    rex = re.compile(r'^([0-9\.]+) = ((.*): )?"?([^"]*)')
+    ret_dict = {}
+    ret_lines = snmp_proc.stdout.readlines()
+
+    if 'No Such Object available on this agent at this OID' in ret_lines[0]:
+        # Would be so much easier to just get snmpwalk return code but
+        # unfortunately it always terminates with code 0
+        return None
+
+    for line in ret_lines:
+        line = line.rstrip()
+        # example row:
+        # .1.3.6.1.2.1.31.1.1.1.6.48 = Counter64: 2861909107355
+        rem = rex.match(line)
+        if rem:
+            reg = rem.groups()
+            res_oid = reg[0]
+            res_value = reg[3]
+            res_array = res_oid.split('.')
+            res_num = int(res_array[-1])
+            if skip:
+                if res_num < skip:
+                    ret_dict[res_num] = res_value
+            else:
+                ret_dict[res_num] = res_value
+
+    return ret_dict
+
+
+def get_switch_data(switch, community):
+    switch_model = get_snmp_value(switch, community, '.1.3.6.1.2.1.1.1.0')
+
+    if switch_model is None:
+        return
+
+    skip = None
+
+    if 'PowerConnect' in switch_model:
+        port_name_dict = get_snmp_table(switch, community, port_name_oid, None)
+        switch_type = 'Dell'
+        trunk_group_tab = get_snmp_table(switch, community, cisco_agg_oid, 0)
+        trunked_ports = set()
+        active_lags = set()
+        for port_idx in trunk_group_tab:
+            if int(trunk_group_tab[port_idx]) > 0:
+                trunked_ports.add(int(port_idx))
+                active_lags.add(int(trunk_group_tab[port_idx]))
+
+    elif 'ProCurve' in switch_model:
+        port_name_dict = get_snmp_table(switch, community, port_name_oid, None)
+        switch_type = 'ProCurve'
+        trunk_group_tab = get_snmp_table(switch, community,
+                                         procurve_trunk_oid, 0)
+
+    elif 'ExtremeXOS' in switch_model:
+        port_name_dict = get_snmp_table(switch, community,
+                                        port_name_oid, 1000000)
+        switch_type = 'Extreme'
+        trunk_group_tab = None
+        skip = 1000000
+
+    port_num = 0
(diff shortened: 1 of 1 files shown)


src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+    'ifInDiscards': '.1.3.6.1.2.1.2.2.1.13',
+    'ifOutDiscards': '.1.3.6.1.2.1.2.2.1.19',
+}
+
+
+def get_snmp_value(host, community, node_name):
+    ''' Function retriving value of given command.
+        In the current version only one row response is handled:
+        return last value retireved via SNMP
+    '''
```
hasegeli: PEP 8 says `"""` for docstrings.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+
+    p = subprocess.Popen(
+        ['snmpwalk', '-On', '-c', community, '-v2c', host, node_name],
```
hasegeli: This could be a `tuple`.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
```
hasegeli: Why are you capturing `stderr`, if you are not going to read it?

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+    )
```
hasegeli: Exit code of this is not checked.  Why don't you just use `subprocess.check_output()`?

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+}
+
+
+def get_snmp_value(host, community, node_name):
+    ''' Function retriving value of given command.
+        In the current version only one row response is handled:
+        return last value retireved via SNMP
+    '''
+
+    p = subprocess.Popen(
```
hasegeli: Single letter variable

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+        ['snmpwalk', '-On', '-c', community, '-v2c', host, node_name],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    resp = p.stdout.read()
+
+    rex = re.compile('^([0-9\.]+) = ((.*): )?"?([^"]*)')
+
+    rem = rex.match(resp)
```
hasegeli: You don't need to `compile()`, if you are not going to reuse the pattern.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+    if rem:
+        reg = rem.groups()
+        return reg[3]
+    else:
+        return None
```
hasegeli: `None` is the implicit return value on Python.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+        stderr=subprocess.PIPE,
+    )
+    resp = p.stdout.read()
+
+    rex = re.compile('^([0-9\.]+) = ((.*): )?"?([^"]*)')
+
+    rem = rex.match(resp)
+    if rem:
+        reg = rem.groups()
+        return reg[3]
```
hasegeli: Why are you capturing all those groups, if you only need the third one?

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+    rex = re.compile('^([0-9\.]+) = ((.*): )?"?([^"]*)')
+
+    ret_dict = {}
+
+    ret_lines = p.stdout.readlines()
+
+    if (
+        ret_lines[0].find('No Such Object available on this agent at this OID')
+        != -1
+    ):
```
hasegeli: `'No Such ...' in  ret_lines[0]`

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+            reg = rem.groups()
+            res_oid = reg[0]
+            res_value = reg[3]
+            res_array = res_oid.split('.')
+            res_num = int(res_array[-1])
+            if skip:
+                if res_num < skip:
+                    ret_dict[res_num] = res_value
+            else:
+                ret_dict[res_num] = res_value
```
hasegeli: It would be nicer to initialise `skip` as `0`.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+        trunk_group_tab = get_snmp_table(switch, community, cisco_agg_oid, 0)
+
+        trunked_ports = set()
+        active_trunks = set()
+        for port_idx in trunk_group_tab:
+            if int(trunk_group_tab[port_idx]) > 0:
+                trunked_ports.add(int(port_idx))
+                active_trunks.add(int(trunk_group_tab[port_idx]))
+
+    if switch_model.find('ProCurve') != -1:
```
hasegeli: `'ProCurve' in switch_model`

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+    # Let's consider that "now" is after all values are
+    # read for this particular switch.
+    now = str(int(time.time()))
+
+    for port_index, port_name in port_name_dict.iteritems():
+        # Skip bad ports. This can be a VLAN on Cisco device.
+        if (
+            port_index not in snmp_tables['bytesIn']
+            or
+            port_index not in snmp_tables['pktsIn']
```
hasegeli: I would write something like `not all(port_index in snmp_tables[k] for k in ('bytesIn', 'pktsIn'))`.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+        elif (
+            (
+                switch_type == 'Dell'
+                and
+                re.match('^(Gi|Te)[0-9]/[0-9]/[0-9]+$', port_name)
+            )
+            or
+            (switch_type == 'ProCurve' and re.match('^[0-9]+$', port_name))
+            or
+            (switch_type == "Extreme")
```
hasegeli: Why only this is double-quoted?

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+            supported_port = False  # dont work on unknown switches
+
+        # This means that this is a supported type of port and we want to
+        # send its counters to Graphite
+        if supported_port:
+            for counter in COUNTERS:
+                graphite += "switches.{}.{}-{}.{} {} {}\n".format(
+                    switch, port_index, port_name,
+                    counter, counters[counter], now
+                )
```
hasegeli: Why don't you just print them?

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+    parser.add_argument(
+        'switches', type=str, nargs='+',
+        help='Hostname of a switch'
+    )
+    return parser.parse_args()
+
+
+def main(args):
+    for switch in args.switches:
+        get_switch_data(switch, args.community)
```
hasegeli: Those two better go on top.  People need to see them first.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+                port_num = int(port_name)
+            elif switch_type == 'Dell':
+                # On Dell blade switches ports have non-contignous numbers,
+                # there is space for extra 4 ports after each 20.
+                port_num += 1
+
+            supported_port = True
+
+        else:
+            supported_port = False  # dont work on unknown switches
```
hasegeli: It is already initialised as `False` on top.

hasegeli: Though, I would just `continue` there instead of this variable.

src/switch_snmp.py
```diff
@@ -0,0 +1,236 @@
...
+    for counter, snmp_oid in COUNTERS.items():
+        snmp_tables[counter] = get_snmp_table(
+            switch, community, snmp_oid, skip
+        )
+
+    # Let's consider that "now" is after all values are
+    # read for this particular switch.
+    now = str(int(time.time()))
+
+    for port_index, port_name in port_name_dict.iteritems():
```
hasegeli: `s/iteritems/items/` for Python 3 compatibility.

src/switch_snmp.py
```diff
@@ -0,0 +1,234 @@
...
+                graphite += "switches.{}.{}-{}.{} {} {}\n".format(
+                    switch, port_index, port_name,
+                    counter, counters[counter], now
+                )
+
+    print(graphite)
+
+
+def parse_args():
+    parser = ArgumentParser()
```
hasegeli: Now almost all scripts accept the `--prefix` argument.  It would be nice, if you can support it too.

src/switch_snmp.py
```diff
@@ -0,0 +1,234 @@
...
+        help='SNMP community'
+    )
+    parser.add_argument(
+        'switches', type=str, nargs='+',
+        help='Hostname of a switch'
+    )
+    return parser.parse_args()
+
+
+def main(args):
```
hasegeli: Now almost all scripts call `parse_args()` inside `main()`.  Lets be consistent on this one too.

src/switch_snmp.py
```diff
@@ -0,0 +1,234 @@
...
+                    counter, counters[counter], now
+                )
+
+    print(graphite)
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument(
+        '-c', '--community', type=str, required=True,
```
hasegeli: `type=str` is unnecessary.

hasegeli: Why do we accept both short and long arguments?  I think only long are fine in this project.  I will add a note to the README about it.

hasegeli: I don't like `--` arguments being "required".  I think `'public'` would be a good default for this.

src/switch_snmp.py
```diff
@@ -0,0 +1,234 @@
...
+
+import re
+import subprocess
+import time
+
+port_name_oid = '.1.3.6.1.2.1.31.1.1.1.1'
+procurve_trunk_oid = '.1.3.6.1.4.1.11.2.14.11.5.1.7.1.3.1.1.8'
+cisco_agg_oid = '.1.2.840.10006.300.43.1.2.1.1.12'
+
+COUNTERS = {
```
hasegeli: It doesn't look like something needs to be a `dict`.  I would go for a tuple of tuples.  Tuple is also more suitable to globals, because of being immutable.

src/switch_snmp.py
```diff
@@ -0,0 +1,234 @@
...
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+
+    rex = re.compile('^([0-9\.]+) = ((.*): )?"?([^"]*)')
+
+    ret_dict = {}
+
+    ret_lines = p.stdout.readlines()
+
```
hasegeli: Why are there empty lines between every other?

src/switch_snmp.py
TuxPowered42: How would I address the values if not in a dict? Just by integer as index in tuple? That's not very nice.

PR #27, opened 2016-12-19
Rework stat_per_process.py

hasegeli:


src/processes.py
```diff
@@ -23,8 +23,6 @@ def main():
     for process_name in get_process_list('/etc/igcollect/stat_per_process.cfg'):
         process_name = process_name.replace('\n', '')
         cpu, mem = get_process_data(process_name)
-        if not cpu or not mem:
-            continue
```
TheFRedFox: What will now happen if there are not present? Is it safe? Is it even possible? ^^

hasegeli: It will emit garbage which is better than silently ignoring the problem.

leoleovich:
Please do not merge it, because it is major change and needs to be supported from other places.

PR #38, opened 2017-05-02
Monitoring of port traffic and cpu usage of switches

TuxPowered42:


src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+import re
+import sys
+
+snmp_engine = SnmpEngine()
+context_data = ContextData()
+
+
+OIDs = {
```
hasegeli: There are more than enough naming styles we have to deal with.  Please do not invent new ones.

src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+    ports_stats(args.prefix, snmp, monitored_ports)
+    cpu_stats(args.prefix, snmp, model)
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('switch', type=str, help='Hostname of a switch')
+    parser.add_argument('--prefix', help='Graphite prefix')
+
+    snmp_mode = parser.add_mutually_exclusive_group(required=True)
```
hasegeli: Cool

src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+
+    monitored_ports = get_monitored_ports(snmp, model)
+    ports_stats(args.prefix, snmp, monitored_ports)
+    cpu_stats(args.prefix, snmp, model)
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('switch', type=str, help='Hostname of a switch')
+    parser.add_argument('--prefix', help='Graphite prefix')
```
hasegeli: Please provide the default.

src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+
+    snmp_mode = parser.add_mutually_exclusive_group(required=True)
+    snmp_mode.add_argument('--community', help='SNMP community')
+    snmp_mode.add_argument('--user', help='SNMPv3 user')
```
hasegeli: `'public'` would be an okay default for `--community`.  I am not sure how the defaults behave with the `mutually_exclusive_group`.

src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+    model = get_snmp_value(snmp, OIDs['switch_model'])
+
+    if 'PowerConnect' in model:
+        return 'powerconnect'
+    elif 'ProCurve' in model:
+        return 'procurve'
+    elif 'ExtremeXOS' in model:
+        return'extreme'
+    else:
+        print('Unknown switch model', file=sys.stderr)
```
hasegeli: You don't need `elif`s and `else` in here.

src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+        lagg_oid = None
+
+    if lagg_oid:
+        return {
+            x: y for x, y
+            in get_snmp_table(snmp, lagg_oid).items()
+            if y != 0
+        }
+
+    return None
```
hasegeli: This can be simplified.

src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+    elif re.match('\A[0-9]+\Z', port_name):
+        # HP normal port
+        return port_name
+    elif re.match('\A[0-9]:[0-9]+\Z', port_name):
+        # Extreme normal port
+        return port_name
+    elif re.match('\A(Gi|Te)[0-9]/[0-9]/[0-9]+\Z', port_name):
+        # Dell normal port
+        return port_name.replace('/', '_')
+    return None
```
hasegeli: You don't need `elif`s.

src/switch.py
```diff
@@ -0,0 +1,275 @@
...
+        # SNMP is slow that we want need to get current time for each counter.
+        template = prefix + '.ports.{}.{} {} ' + str(int(time()))
+        table = get_snmp_table(snmp, oid)
+        for port_idx, port_name in ports.items():
+            print(template.format(port_name, counter, table[port_idx]))
+
+
+def cpu_stats(prefix, snmp, model):
+    cpu_usage = get_snmp_value(snmp, CPU_OIDs[model])
+    template = prefix + '.cpu {} ' + str(int(time()))
```
hasegeli: It would be nicer to call `time()` only once.

src/switch.py
TuxPowered42: Does not seem to work: "switch.py: error: one of the arguments --community --user is required"

src/switch.py
TuxPowered42: I made them all upper case, although is is for plural and you know that I don't agree with that notation.

src/switch.py
TuxPowered42: To be honest I am not sure how to handle it at all to make it compatible with our usual prefixes. When run from igcollect puppet helper prefix would be "switches.$SWITCH" for every switch cron job. But what would be the default then?

src/switch.py
TuxPowered42: Done. Sorry, I was already on process of simplifying this but somehow I missed it.

src/switch.py
TuxPowered42: True.

src/switch.py
hasegeli: I haven't noticed that you are assigning a default later.  I guess it is okay than.  All other scripts simply setting their name as the default prefix.  `nginx.py` would print metrics like `nginx.xxx`.  We could do something more complicated by getting the hostname, but I preferred not to do so, as this would most likely be different for everybody.

src/switch.py
```diff
@@ -0,0 +1,296 @@
...
+    """ Recognize model of switch from SNMP MIB-2 sysDescr """
+
+    model = get_snmp_value(snmp, OIDS['switch_model'])
+
+    if 'PowerConnect' in model:
+        return 'powerconnect'
+    elif 'ProCurve' in model:
+        return 'procurve'
+    elif 'ExtremeXOS' in model:
+        return'extreme'
```
hasegeli: You don'y need `elif` in here either.

src/switch.py
```diff
@@ -0,0 +1,296 @@
...
+
+    print('Unknown switch model', file=sys.stderr)
```
hasegeli: I would prefer raising an `Exception` to `print()`ing from utility functions.

src/switch.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def get_laggs(snmp, model):
+    """ Get only those LAGGs which have members """
+
+    if model in ['powerconnect', 'procurve']:
+        return {
+            x: y for x, y
+            in get_snmp_table(snmp, LAGG_OIDS[model]).items()
+            if y != 0
```
hasegeli: Can it be `< 0`?

src/switch.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def cpu_stats(prefix, snmp, model):
+    """ Print graphite-compatible stats of switch CPU
+
+        We use single OID which should percentage of CPU time used.
+    """
+
+    cpu_usage = get_snmp_value(snmp, CPU_OIDS[model])
+    template = prefix + '.cpu {} ' + str(int(time()))
```
hasegeli: I would save the time once and reuse it.

PR #39, opened 2017-05-02
Drop redundant words from resource script names

hasegeli:


src/linux_disk.py
```diff
@@ -11,7 +11,7 @@
 
 def parse_args():
     parser = ArgumentParser()
-    parser.add_argument('--prefix', default='disk_io')
+    parser.add_argument('--prefix', default='disk')
```
TuxPowered42: I approve this name because we are using such metric name anyway. Also there is not many other "disk" things.

PR #45, opened 2017-05-05
Add java_jvm to collect jvm metrics using jstat

kofrezo:


src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+        help='Number of samples before terminating'
+    )
+    parser.add_argument(
+        '--jvm-install-path',
+        help='JVM installation path e.g. /usr/lib/jvm/jdk-8-oracle-x64',
+        required=True
+    )
+    parser.add_argument(
+        '--prefix',
+        default='jvm.jstat',
```
hasegeli: The commit message says "java_jvm".  The script is called "java_jstat".  And the default prefix is "jvm.jstat".  Would you stick with one of them?

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+    start = int(time())
+    output = check_output(command, shell=True)
+    rows = output.splitlines()
+    header = rows.pop(0)
+
+    current_interval = 1
+    for row in rows:
+        columns = row.split()
+        h_columns = header.split()
+        for h_index in range(0, len(h_columns)):
```
hasegeli: You can `enumerate(h_columns)` or `zip(columns, h_columns)` or even both.

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+
+    start = int(time())
+    output = check_output(command, shell=True)
+    rows = output.splitlines()
+    header = rows.pop(0)
+
+    current_interval = 1
+    for row in rows:
+        columns = row.split()
+        h_columns = header.split()
```
hasegeli: I don't like this naming style.  How about simply calling it `column_names`?

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+    start = int(time())
+    output = check_output(command, shell=True)
+    rows = output.splitlines()
+    header = rows.pop(0)
+
+    current_interval = 1
+    for row in rows:
+        columns = row.split()
```
hasegeli: I think `cells` would be more appropriate.

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+        count
+    )
+
+    start = int(time())
+    output = check_output(command, shell=True)
+    rows = output.splitlines()
+    header = rows.pop(0)
+
+    current_interval = 1
+    for row in rows:
```
hasegeli: `enumerate(rows, start=1)`

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+        help='jstat option to collect e.g. gcutil or all if specified'
+    )
+    parser.add_argument(
+        '--vmid',
+        nargs='+',
+        help='Java VM id to gather metrics from, check jps tool'
+    )
+    parser.add_argument(
+        '--interval',
+        default='1',
```
hasegeli: Why not `int`?

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+    )
+    parser.add_argument(
+        '--count',
+        default=4,
+        help='Number of samples before terminating'
+    )
+    parser.add_argument(
+        '--jvm-install-path',
+        help='JVM installation path e.g. /usr/lib/jvm/jdk-8-oracle-x64',
+        required=True
```
hasegeli: I don't like required "optional argumens".  Can't we choose a reasonable default path?

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+#
+# Copyright (c) 2017, InnoGames GmbH
+
+from time import time
+from argparse import ArgumentParser
+from subprocess import check_output
+
+
+def main():
+    args = vars(parse_args())
```
hasegeli: You could keep it as an `object`.  It looks like you don't need it as a `dict` later.

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+
+    if args['option'] is None:
```
hasegeli: I am not sure, if `argparse` would give you `None` or an empty `list`.

src/java_jstat
```diff
@@ -0,0 +1,125 @@
...
+def get_options(jvm_install_path):
+    """Get all available jstat options -> list
+
+    Arguments:
+        jvm_install_path -- absolute path to jvm installation
+    """
+
+    command = '{}/bin/jstat -options'.format(jvm_install_path)
+    for option in check_output(command, shell=True).splitlines():
+        yield option[1:]
```
hasegeli: It would be nice to comment why the first is omitted.

src/java_jstat
kofrezo: None i tested

src/java_jstat
kofrezo: Ah yes I had 1s before but then recognized that makes no sense for graphite because it can only handle timestamps with full seconds - will change.

src/java_jstat
kofrezo: If you have an idea I could not find a platform independent reasonable way to find out absolute path an java version.

src/java_jstat
kofrezo: Sure

src/java_jstat
kofrezo: Will do

src/java_jstat
Phantuman: Could you at least on Linux try to find the path?

PR #47, opened 2017-05-17
Convert rss to bytes

leoleovich:
ps gives you output in Kilobytes, which is pretty bad, because you need to perform extra magic within graphite to convert it to other units.

### DIFF
--- a/src/processes.py
+++ b/src/processes.py
@@ -32,7 +32,7 @@ def main():
                 print(template.format(command, 'cpu_usage', process.pcpu))
                 print(template.format(command, 'mem_usage', process.pmem))
                 print(template.format(command, 'etimes', process.etimes))
-                print(template.format(command, 'rss', process.rss))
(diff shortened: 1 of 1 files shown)


src/processes.py
```diff
@@ -32,7 +32,7 @@ def main():
                 print(template.format(command, 'cpu_usage', process.pcpu))
                 print(template.format(command, 'mem_usage', process.pmem))
                 print(template.format(command, 'etimes', process.etimes))
-                print(template.format(command, 'rss', process.rss))
+                print(template.format(command, 'rss', int(process.rss) * 1024))
```
leoleovich: @TuxPowered42 actually there is no moment of parsing.
https://github.com/innogames/igcollect/blob/ab9d42178de187e1189d158d27d822edc843f2fd/src/processes.py#L46-L47

Only marshalling.

PR #66, opened 2017-07-25
Add support for Dell MXL switch

TuxPowered42:


src/switch.py
```diff
@@ -262,6 +266,14 @@ def standarize_portname(port_name):
     if re.match('\A(Gi|Te)[0-9]/[0-9]/[0-9]+\Z', port_name):
```
TheFRedFox: `[0-9]` could be replaced with `\d` :wink:

PR #69, opened 2017-10-25
Add mysql_query script

v0devil:


src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
+#
+# igcollect - PostgreSQL query results
```
hasegeli: "MySQL"

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+#
+# This script executes a single query and prints the results.  The columns
+# returned by the query are going to be appended to the given prefix.
+# The query must return numeric values.
+#
+# Copyright (c) 2016, InnoGames GmbH
```
hasegeli: "2017"

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+#
+from __future__ import print_function
+
+import MySQLdb
```
hasegeli: Please use `mysql.connector` instead from Oracle.  MySQLdb is not available for Python 3, and I don't think gets much maintenance.

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+
+from argparse import ArgumentParser
+from time import time
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='mysql_query')
+    parser.add_argument('--dbname', default='mysql')
```
hasegeli: `default='test'` might me better.  I don't think MySQL has a database called `mysql` by default.

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+    parser.add_argument(
+        '--query',
+        required=True,
+        action='append',
+        dest='queries', )
```
hasegeli: I would prefer to have `)` on a new line same level as `(`.

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+
+def main():
+    """The main program"""
+    args = parse_args()
+    items = []
+    now = str(int(time()))
+
+    db = MySQLdb.connect(
+        user='root',
+        host="localhost",
```
hasegeli: These 2 should also be arguments.

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def get_column_data(rows):
+    if len(rows) > 1:
+        raise Exception('Multiple rows')
+    for row in rows:
+        return row.items()
+
+
+def get_row_data(key_column, rows):
```
hasegeli: I would put `rows` as the first argument to make the function signature compatible with the one above.

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+    first_row = rows[0]
+    if key_column not in first_row:
+        raise Exception('Key column is not there')
+    if len(first_row) == 1:
```
hasegeli: I am not sure whether it can happen, but `< 2` would be a safer check.

src/mysql_query.py
```diff
@@ -0,0 +1,81 @@
...
+        user='root',
+        host="localhost",
+        db=args.dbname,
+        read_default_file='/etc/mysql/my.cnf', )
+    cur = db.cursor(MySQLdb.cursors.DictCursor)
+    for query in args.queries:
+        cur.execute(query)
+        if not cur.rowcount:
+            raise Exception('No result')
+        else:
```
hasegeli: Unnecessary `else`

src/mysql_query.py
elmatze79: the default databases are:
mysql
performance_schema
information_schema
It is highly advised to remove the test database after install

PR #70, opened 2017-10-26
Add haproxy

v0devil:


src/haproxy.py
```diff
@@ -0,0 +1,51 @@
...
+        for px_key, px_value in px_d.items():
+            for sv_key, sv_value in px_value.items():
+                for metric_name, metric_value in sv_value.items():
+                    print(
+                        template.format(px_key, sv_key, metric_name,
+                                        metric_value))
+
+
+if __name__ == '__main__':
+    main()
```
hasegeli: New line please

src/haproxy.py
```diff
@@ -0,0 +1,51 @@
...
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='haproxy')
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    haproxy_info = Popen(
+        ('echo "show stat" | socat /var/run/haproxy.stat stdio'),
```
hasegeli: It is not cool to run shell commands for things we can do in Python.  You can use `stdin=PIPE`, and `proc.communicate()`.

hasegeli: Maybe we can do this `socat` on Python as well to avoid calling any shell command.

src/haproxy.py
```diff
@@ -0,0 +1,51 @@
...
+        shell=True,
+        stdout=PIPE).stdout.read()
+    haproxy_info = haproxy_info.splitlines()
+    header = haproxy_info[0].replace(" ", "_").split(',')
+    data_overall = []
+    for n in xrange(1, len(haproxy_info) - 1):
```
hasegeli: `xrange()` is no more on Python 3.  We have Python 3 only servers.  Please test your code on Python 3.

src/haproxy.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def main():
+    args = parse_args()
+    haproxy_info = Popen(
+        ('echo "show stat" | socat /var/run/haproxy.stat stdio'),
+        shell=True,
+        stdout=PIPE).stdout.read()
+    haproxy_info = haproxy_info.splitlines()
+    header = haproxy_info[0].replace(" ", "_").split(',')
```
hasegeli: Let's not mix double and single quotes randomly.

src/haproxy.py
```diff
@@ -0,0 +1,51 @@
...
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    haproxy_info = Popen(
+        ('echo "show stat" | socat /var/run/haproxy.stat stdio'),
+        shell=True,
+        stdout=PIPE).stdout.read()
+    haproxy_info = haproxy_info.splitlines()
```
hasegeli: It would have been more efficient to read lines one by one compared to reading them all and splitting.

src/haproxy.py
hasegeli: Better use `enumerate()` there

src/haproxy.py
```diff
@@ -0,0 +1,51 @@
...
+        service_data = haproxy_info[n].split(',')
+        i = 0
+        data = {}  # gather data for every proxy/service
+        for h in header:
+            data[h] = service_data[i]
+            i += 1
+        pxname = data['#_pxname']
+        svname = data['svname']  # pxname - proxy name; svname - service name
+        del data['#_pxname']
+        del data['svname']
```
hasegeli: You could `data.pop()` instead of reading and deleting them.

src/haproxy.py
```diff
@@ -0,0 +1,51 @@
...
+        shell=True,
+        stdout=PIPE).stdout.read()
+    haproxy_info = haproxy_info.splitlines()
+    header = haproxy_info[0].replace(" ", "_").split(',')
+    data_overall = []
+    for n in xrange(1, len(haproxy_info) - 1):
+        service_data = haproxy_info[n].split(',')
+        i = 0
+        data = {}  # gather data for every proxy/service
+        for h in header:
```
hasegeli: `enumerate()` here too

src/haproxy.py
v0devil: https://stackoverflow.com/questions/17226912/why-does-simple-echo-in-subprocess-not-working

src/haproxy.py
v0devil: i am creating here a lines

src/haproxy.py
```diff
@@ -0,0 +1,58 @@
...
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    haproxy_info = read_ha_proxy_stats('/var/run/haproxy.stat')
+    haproxy_info = haproxy_info.splitlines()
+    header = haproxy_info[0].replace(' ', '_').split(',')
+    data_overall = []
+    for n in range(1, len(haproxy_info) - 1):
```
hasegeli: `for line in haproxy_info[1:-1]`

src/haproxy.py
```diff
@@ -0,0 +1,58 @@
...
+                        template.format(px_key, sv_key, metric_name,
+                                        metric_value))
+
+
+def read_ha_proxy_stats(haproxy_stat):
+    conn = socket(AF_UNIX, SOCK_STREAM)
+    try:
+        conn.connect(haproxy_stat)
+        conn.sendall('show stat')
+        return conn.recv(4096)
```
hasegeli: Are you sure this is always enough?  Please add a comment about it.

src/haproxy.py
```diff
@@ -0,0 +1,58 @@
...
+    haproxy_info = haproxy_info.splitlines()
+    header = haproxy_info[0].replace(' ', '_').split(',')
+    data_overall = []
+    for n in range(1, len(haproxy_info) - 1):
+        service_data = haproxy_info[n].split(',')
+        i = 0
+        data = {}  # gather data for every proxy/service
+        for h in header:
+            data[h] = service_data[i]
+            i += 1
```
hasegeli: `data = dict(zip(header, service_data))`

src/haproxy.py
```diff
@@ -0,0 +1,58 @@
...
+        pxname = data.pop('#_pxname') # pxname - proxy name; svname - service name
+        svname = data.pop('svname')
+        data_overall.append({pxname: {svname: data}})
```
hasegeli: Why don't you just print it in here?

src/haproxy.py
```diff
@@ -0,0 +1,61 @@
...
+    conn = socket(AF_UNIX, SOCK_STREAM)
+    try:
+        conn.connect(haproxy_stats_socket)
+        conn.sendall('show stat\r\n')
+        data = conn.recv(BUFFER_SIZE)
+        if len(data) == BUFFER_SIZE:
+            while 1:
+                try:
+                    data += conn.recv(BUFFER_SIZE, socket.MSG_DONTWAIT)
+                except:
```
hasegeli: It is a bad idea to `except` anything.  You could even get syntax errors in here.  What are you really expecting?

src/haproxy.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def read_ha_proxy_stats(haproxy_stats_socket):
+    conn = socket(AF_UNIX, SOCK_STREAM)
+    try:
+        conn.connect(haproxy_stats_socket)
+        conn.sendall('show stat\r\n')
+        data = conn.recv(BUFFER_SIZE)
+        if len(data) == BUFFER_SIZE:
+            while 1:
```
hasegeli: Wouldn't it better to say something like `while len(data) == BUFFER_SIZE` and concatenate them to each other later?  Abusing endless loops is considered bad for readability.

src/haproxy.py
```diff
@@ -0,0 +1,61 @@
...
+    args = parse_args()
+    template = args.prefix + '.{}.{}.{} {} ' + str(int(time()))
+    haproxy_info = read_ha_proxy_stats(args.haproxy_stats_socket)
+    haproxy_info = haproxy_info.splitlines()
+    header = haproxy_info[0].replace(' ', '_').split(',')
+    data_overall = []
+    for line in haproxy_info[1:-1]:
+        service_data = line.split(',')
+        data = dict(zip(header,
+                        service_data))  # gather data for every proxy/service
```
hasegeli: This is pretty weird formatting in here.  You can write:
```python
        # Gather data for every proxy/service
        data = dict(zip(header, service_data))
```

src/haproxy.py
```diff
@@ -0,0 +1,61 @@
...
+        pxname = data.pop(
+            '#_pxname')  # pxname - proxy name; svname - service name
```
hasegeli: Same in here

PR #71, opened 2017-10-27
Add redis_keys

v0devil:


src/redis_keys.py
```diff
@@ -0,0 +1,35 @@
...
+    parser.add_argument('--command', default='llen')
+    parser.add_argument('--keys', default='*queue*')
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+
+    template = args.prefix + '.{}.{} {} ' + str(int(time()))
+    queues = []
```
hasegeli: Not used

src/redis_keys.py
```diff
@@ -0,0 +1,35 @@
...
+    output = Popen(('redis-cli', 'keys', args.keys), stdout=PIPE).stdout.read()
```
hasegeli: You used a `list` on the `Popen` below, and `tuple` in here.  Let's keep things consistent and use ` list` in here too.

hasegeli: I don't think `.stdout.read()` is appropriate way of using `Popen`.  Please use `.communicate()`.

src/redis_keys.py
```diff
@@ -0,0 +1,35 @@
...
+    for queue in output.splitlines():
+        data = Popen(
+            ['redis-cli', args.command, queue],
+            stdout=PIPE).stdout.read().strip('\n')
```
hasegeli: This script would be more efficient by using the appropriate Python Redis driver and keeping connection open.

hasegeli: By the way, you don't need to give arguments to `.strip()`.

PR #74, opened 2017-11-01
Fix haproxy for python3

v0devil:


src/haproxy.py
```diff
@@ -41,7 +41,7 @@ def read_ha_proxy_stats(haproxy_stats_socket):
     conn = socket(AF_UNIX, SOCK_STREAM)
     try:
         conn.connect(haproxy_stats_socket)
-        conn.sendall('show stat\r\n')
+        conn.sendall('show stat\r\n'.encode('utf-8'))
```
hasegeli: `b'show stat\r\n'`

src/haproxy.py
```diff
@@ -55,3 +55,4 @@ def read_ha_proxy_stats(haproxy_stats_socket):
 
 if __name__ == '__main__':
     main()
+
```
hasegeli: Extra line

src/haproxy.py
```diff
@@ -25,7 +25,7 @@ def parse_args():
 def main():
     args = parse_args()
     template = args.prefix + '.{}.{}.{} {} ' + str(int(time()))
-    haproxy_info = read_ha_proxy_stats(args.haproxy_stats_socket)
+    haproxy_info = read_ha_proxy_stats(args.haproxy_stats_socket).decode('utf-8')
```
hasegeli: I don't think this is unicode.  `.decode()` without argument should be sufficient.

src/haproxy.py
```diff
@@ -54,4 +54,4 @@ def read_ha_proxy_stats(haproxy_stats_socket):
 
 
 if __name__ == '__main__':
-    main()
+    main()
```
hasegeli: Now it removes the line ending that was there.

PR #77, opened 2017-12-18
Add plugin to request data from JSON url endpoint

v0devil:


src/url_json.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# Copyright (c) 2017, InnoGames GmbH
+#
+
+from argparse import ArgumentParser
+from time import time
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
```
hasegeli: Do we really need to support Python 2?

src/url_json.py
```diff
@@ -0,0 +1,38 @@
...
+import json
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='data')
+    parser.add_argument('--url', default='http://localhost/')
+    parser.add_argument('--keys', default='key1,key2')
```
hasegeli: It would better be a multi argument with `action='append'`.  You would then get the keys in a `list`.

src/url_json.py
```diff
@@ -0,0 +1,38 @@
...
+try:
+    from urllib.request import urlopen
+except ImportError:
+    from urllib2 import urlopen
+import json
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='data')
```
hasegeli: All other plugins default the prefix to their name, so `url_json`.

PR #80, opened 2017-12-19
Add plugin to parse values from log file

v0devil:


src/logfile_values.py
```diff
@@ -0,0 +1,57 @@
...
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='logfile_values')
+    parser.add_argument('--file', default='1.log')
+    parser.add_argument(
+        '-n',
+        '--name',
+        help='Graphite metric name',
+        default=None,
```
hasegeli: This doesn't look useful.  The default would be `[]` when `action='append'`.

src/logfile_values.py
```diff
@@ -0,0 +1,57 @@
...
+"""
+
+from argparse import ArgumentParser
+from time import time
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='logfile_values')
+    parser.add_argument('--file', default='1.log')
```
hasegeli: This is a strange default.  How about something like `/var/log/messages`?

src/logfile_values.py
```diff
@@ -0,0 +1,57 @@
...
+    parser.add_argument(
+        '-n',
```
hasegeli: Do we really need the short variants?

src/logfile_values.py
```diff
@@ -0,0 +1,57 @@
...
+    parser.add_argument(
+        '-c',
+        '--column',
+        help='Number of the column with metric value',
+        default=None,
+        action='append',
+        dest='columns', )
+    args = parser.parse_args()
+    if len([x for x in (args.names) if x is not None]) != len(
+        [x for x in (args.columns) if x is not None]):
```
hasegeli: This would be `len(args.name) == len(args.columns)` without the `default`s.

src/logfile_values.py
```diff
@@ -0,0 +1,57 @@
...
+        default=None,
+        action='append',
+        dest='names', )
+    parser.add_argument(
+        '-c',
+        '--column',
+        help='Number of the column with metric value',
+        default=None,
+        action='append',
+        dest='columns', )
```
hasegeli: You can add there `type=int` to avoid casting later.

src/logfile_values.py
```diff
@@ -0,0 +1,57 @@
...
+    args = parser.parse_args()
```
hasegeli: I don't like dependent command like arguments.  I would design it something like this:

```python
class Metric:
    def __init__(self, arg):
        if ':' not in arg:
            raise ArgumentTypeError('Argument must have ":"')
        self.name, column = arg.split(':', 1)
        if not column.isdecimal():
            raise ArgumentTypeError('Column must be a number')
        self.column = int(column)
```

```python
    parser.add_argument('metric', type=Metric, action='append')
```

`ArgumentParser` lets you use simple functions as types as well, but I think class is better for this script.  We can extend it later to get the metrics from lines other than the last one.

src/logfile_values.py
TheFRedFox: Generally `[x for x in (args.names) if x is not None]` can be replaced with:
`filter(None, args.names)`

PR #81, opened 2017-12-20
Add nargs=+ for metric argument in logfile_values

v0devil:


src/logfile_values.py
```diff
@@ -30,7 +30,7 @@ def parse_args():
     parser = ArgumentParser()
     parser.add_argument('--prefix', default='logfile_values')
     parser.add_argument('--file', default='/var/log/messages')
-    parser.add_argument('--metric', type=Metric, action='append')
+    parser.add_argument('--metric', type=Metric, action='append', nargs='+')
```
hasegeli: You don't need `action='append'` together with `nargs='+'`.

PR #84, opened 2018-01-08
Gentime log

v0devil:
Add experimental log parser.


src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+            raise ArgumentTypeError('Argument must have ":"')
+        parts = arg.split(':')
+        if len(parts) != 4 and len(parts) != 2:
+            raise ValueError('Wring number of options')
+
+        # Accepts 2 or 4 options. If 2 then function and period are NULL
+        parts += [None] * (4 - len(parts))
+        self.name, column, function, period = parts
+        if period:
+            pattern = re.compile("^\d+[A-Za-z]$")
```
hasegeli: We should compile it outside the constructor for efficiency.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+        self.last_value = 0
+
+    def get_timeshift(self):
+        if self.period:
+            value = int(self.period[:-1])
+            unit = self.period[-1].lower()
+            now = int(time.time())
+            if unit == 's':
+                return timedelta(seconds=value).total_seconds()
+            elif unit == 'm':
```
hasegeli: No need for `elif`

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+                return timedelta(minutes=value).total_seconds()
+            elif unit == 'h':
+                return timedelta(hours=value).total_seconds()
+            elif unit == 'd':
+                return timedelta(days=value).total_seconds()
+        else:
```
hasegeli: No need for `else`

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+            return 0
+
+    def get_median(self):
+        l = sorted(self.values)
+        i = len(l)
+        if not i % 2:
+            return (l[(i // 2) - 1] + l[i // 2]) / 2
```
hasegeli: `sorted()` doesn't return a `list` on Python 3.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+        return int(l[i // 2])
+
+    def get_sum(self):
+        return int(sum(self.values))
```
hasegeli: Why do you cast it to `int`?

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+            return self.get_min()
+        elif f == 'max':
+            return self.get_max()
+        elif f == 'count':
+            return self.get_count()
+        elif f == 'last' or not f:
+            return self.get_last_value()
+        else:
+            raise ArgumentTypeError(
+                'Wrong function. Possible functions: blablabla')
```
hasegeli: We can make this selection more elegant by something like `return getattr(self, 'get_' + self.function)`.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='logfile_values')
+    parser.add_argument('--file', default='/var/log/messages')
+    parser.add_argument('--metric', type=Metric, nargs='+')
+    parser.add_argument('--time_format', default='%Y-%m-%dT%H:%M:%S')
```
hasegeli: Is the time zone also recognised?

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+                    lines[-1] += segment
+                else:
+                    yield segment
+            segment = lines[0]
+            for index in range(len(lines) - 1, 0, -1):
+                if len(lines[index]):
+                    yield lines[index]
+        # Don't yield None if the file was empty
+        if segment is not None:
+            yield segment
```
hasegeli: Last 3 lines can go outside `with:`.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+        segment = None
+        offset = 0
+        fh.seek(0, os.SEEK_END)
+        file_size = remaining_size = fh.tell()
+        while remaining_size > 0:
+            offset = min(file_size, offset + buf_size)
+            fh.seek(file_size - offset)
+            buffer = fh.read(min(remaining_size, buf_size))
+            remaining_size -= buf_size
+            lines = buffer.split('\n')
```
hasegeli: `splitlines()` is safer.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def convert_to_timestamp(time_str, time_format):
+    try:
+        timestamp = int(
+            time.mktime(time.strptime(time_str.split("+")[0], time_format)))
+    except ValueError:
+        try:
+            timestamp = int(
+                time_str)
```
hasegeli: No need for a new line.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+                time.mktime(
+                    time.strptime('-'.join(time_str.split("-")[:-1]),
+                                  time_format)))
+        pass
+    return int(timestamp)
+
+
+def main():
+    args = parse_args()
+    file = args.file
```
hasegeli: Why don't we just use `args.file`?

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+                                fields = line.split()
+                                fields[0] = convert_to_timestamp(
+                                    fields[0], args.time_format)
+                                if fields[0] > (int(time.time()) -
+                                                metric.get_timeshift()):
+                                    metric.values.append(
+                                        int(fields[metric.column]))
+
+    for metric in args.metric:
+        print(template.format(metric.name, metric.get_metric_value()))
```
hasegeli: We need to split this function.

src/logfile_values_experimental.py
elmatze79: Why? Shouldn't the use of `elif` omit the if check, when `unit == s`?

hasegeli: There are `return`s.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+from argparse import ArgumentParser, ArgumentTypeError
+from datetime import timedelta
+
+
+class Metric:
+    def __init__(self, arg):
+        if ':' not in arg:
+            raise ArgumentTypeError('Argument must have ":"')
+        parts = arg.split(':')
+        if len(parts) != 4 and len(parts) != 2:
```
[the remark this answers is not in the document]
v0devil: We check if X == 2 (A) or X == 4 (B) otherwise return error
A or B -> !(A or B) -> A! and B! ?

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,221 @@
...
+            raise ValueError('Wring number of options')
+
+        # Accepts 2 or 4 options. If 2 then function and period are NULL
+        parts += [None] * (4 - len(parts))
+        self.name, column, function, period = parts
```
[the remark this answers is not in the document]
v0devil: I wanted to make 2 remain variables (funtion and period) = None if user entered only 2 parts.

src/logfile_values_experimental.py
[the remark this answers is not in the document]
hasegeli: True

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+        self.last_value = 0
+        self.now = int(time.time())
+
+    def get_timeshift(self):
+        if self.period:
+            value = int(self.period[:-1])
+            unit = self.period[-1].lower()
+            if unit == 's':
+                return timedelta(seconds=value).total_seconds()
+            if unit == 'm':
```
hasegeli: `m` is not proper abbreviation for "minute".  It is reserved for "metre" on SI standard.  `min` would be the proper one.

src/logfile_values_experimental.py
hasegeli: `-` is preferred over `_` on the command line options.  You can call it `--time-format`.  `ArgumentParser` would still make the value available as `time_format`.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='logfile_values')
+    parser.add_argument('--file', default='/var/log/messages')
+    parser.add_argument('--metric', type=Metric, nargs='+')
+    parser.add_argument('--time_format', default='%Y-%m-%dT%H:%M:%S')
+    parser.add_argument('--arch', action='store_true')
+    parser.add_argument('--debug', '-d', action='store_true')
+    parser.add_argument('--timezone', default='Europe/Berlin')
```
hasegeli: We can also call this `--time-zone` is it is written separated in English.  `UTC` would make a better default.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+    return parser.parse_args()
+
+
+def yield_metric_value(line, metric, time_format):
+    fields = line.split()
+    fields[0] = convert_to_timestamp(fields[0], time_format)
+    if fields[0] > (metric.now - metric.get_timeshift()):
```
hasegeli: Extra parantheses

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+        return fields[metric.column]
+    else:
+        return False
```
hasegeli: I prefer functions to return either True/False, or something/None.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+            if segment is not None:
+                # if the previous chunk starts right from the beginning of line
+                # do not concact the segment to the last line of new chunk
+                # instead, yield the segment first
+                if buffer[-1] is not '\n':
+                    lines[-1] += segment
+                else:
+                    yield yield_metric_value(segment, metric, time_format)
+            segment = lines[0]
+            for index in range(len(lines) - 1, 0, -1):
```
hasegeli: I would write this as `reversed(range(len(lines)))`.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def main():
+    args = parse_args()
+    if args.debug:
+        logging.getLogger().setLevel(logging.DEBUG)
+        logging.getLogger().addHandler(logging.StreamHandler())
+
+    if args.timezone:
+        os.environ['TZ'] = args.timezone
```
hasegeli: Why?

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+        # time.tzset()
+
+    for metric in args.metric:
+        template = args.prefix + '.{} {} ' + str(metric.now)
+        file_was_readed = True
+        # Read from the end of file until the timestamp is satisfying conditions
+        data = read_logfile_reverse(args.file, args.time_format, metric)
```
hasegeli: It is not nice to pass the `time_format` and `metric` to this function.  We should get the lines and do the analyses in here.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,224 @@
...
+    args = parse_args()
+    if args.debug:
+        logging.getLogger().setLevel(logging.DEBUG)
+        logging.getLogger().addHandler(logging.StreamHandler())
+
+    if args.timezone:
+        os.environ['TZ'] = args.timezone
+        # time.tzset()
+
+    for metric in args.metric:
```
hasegeli: This is reading the file for every metric.  Reading the file once to execute the checks on is would be better.

v0devil: But we do not read a whole file and metrics can have a different time periods

src/logfile_values_experimental.py
v0devil: then how i ll understand that file was or not read completely

src/logfile_values_experimental.py
v0devil: Need to exclude zero

src/logfile_values_experimental.py
v0devil: Actually, you are right, if we run script with some period metrics cannot have different aggregation time

src/logfile_values_experimental.py
hasegeli: You can iterate the generator as long as you like.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+        self.period = period
+        # Here is container for metric values
+        self.values = []
+        self.last_value = 0
+        self.now = int(time.time())
+
+    def get_timeshift(self):
+        if self.period:
+            value = int(self.period[:-1])
+            unit = self.period[-1].lower()
```
hasegeli: This wouldn't work now as the unit is not always 1 character.  I am sorry for causing more work to you.  Standards are something we never sacrifice.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def get_last_value(self):
+        return self.last_value
+
+    def get_metric_value(self):
+        if self.function:
+            if 'count_' in self.function:
+                return int(self.get_count(int(self.function.split('_')[1])))
+            return int(getattr(self, 'get_' + self.function)())
+        else:
```
hasegeli: No need to `else:`.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+    parser.add_argument('--file', default='/var/log/messages')
+    parser.add_argument('--metric', type=Metric, nargs='+')
+    parser.add_argument('--time-format', default='%Y-%m-%dT%H:%M:%S')
+    parser.add_argument('--arch', action='store_true')
+    parser.add_argument('--debug', '-d', action='store_true')
+    parser.add_argument('--time-zone', default='UTC')
+    return parser.parse_args()
+
+
+def yield_metrics_values(line, metrics, time_format):
```
hasegeli: This doesn't seem to `yield` but `return`.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+    values = []
+    fields = line.split()
+    fields[0] = convert_to_timestamp(fields[0], time_format)
+    for metric in metrics:
+        if fields[0] > metric.now - metric.get_timeshift():
+            values.append(int(fields[metric.column]))
+        else:
+            return None
+    return values
```
hasegeli: What is the difference between returning an empty `list` or `None`?

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+
+def main():
+    args = parse_args()
+    if args.debug:
+        logging.getLogger().setLevel(logging.DEBUG)
+        logging.getLogger().addHandler(logging.StreamHandler())
+
+    if args.time_zone:
+        os.environ['TZ'] = args.time_zone
+        # time.tzset()
```
hasegeli: Should it be enabled?

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+                                               time_format)
+
+    if segment is not None:
+        yield yield_metrics_values(segment, metrics, time_format)
+
+
+def convert_to_timestamp(time_str, time_format):
+    try:
+        timestamp = int(
+            time.mktime(time.strptime(time_str.split("+")[0], time_format)))
```
hasegeli: You used single quotes on other places.  Let's be consistent.

hasegeli: `split('+', 1)` if you don't need the elements after the first one.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+    data = read_logfile_reverse(
+        args.file, args.time_format, metrics=args.metric)
+    for log_value in data:
+        if log_value:
+            for index, value in enumerate(log_value):
+                args.metric[index].values.append(value)
+        # Stop reading if next line has bigger timestamp as a required
+        else:
+            file_was_readed = log_value
+            break
```
hasegeli: Checking for `if not log_value: break` before would be cleaner.

src/logfile_values_experimental.py
```diff
@@ -0,0 +1,244 @@
...
+            time.mktime(time.strptime(time_str.split("+")[0], time_format)))
+    except ValueError:
+        try:
+            timestamp = int(time_str)
+        except ValueError:
+            timestamp = int(
+                time.mktime(
+                    time.strptime('-'.join(time_str.split("-")[:-1]),
+                                  time_format)))
+        pass
```
hasegeli: Leftover

src/logfile_values_experimental.py
v0devil: Need it, cuz idea is when user does not point function then we just return the last value for current metric

src/logfile_values_experimental.py
v0devil: You sure it`s anyway returning list :-/

src/logfile_values_experimental.py
hasegeli: You don't need the `else:` part because the previous line `return`s.

src/logfile_values_experimental.py
hasegeli: Sure.  The second argument of `split()` is to tell Python to stop after some occurrences.  It is good practice to specify it when you are interested in only some elements in the beginning.

src/logfile_values.py
```diff
@@ -7,42 +7,298 @@
...
-        self.column = int(column)
+        parts = arg.split(':')
+        if len(parts) != 4 and len(parts) != 2:
+            raise ValueError('Wrong number of options')
+
+        # Accepts 2 or 4 options. If 2 then function and period are NULL
+        parts += [None] * (4 - len(parts))
+        self.name, column, function, period = parts
+        if period:
+            pattern = re.compile("^\d+[A-Za-z]{1,3}$")
```
hasegeli: Let's stick with the single quotes, not double quotes on some places.

src/logfile_values.py
```diff
@@ -7,42 +7,298 @@
...
+            value = int(" ".join(re.findall("\d+", self.period.lower())))
+            unit = " ".join(re.findall("[a-z]+", self.period.lower()))
+            if unit == 's':
+                return timedelta(seconds=value).total_seconds()
+            if unit == 'min':
+                return timedelta(minutes=value).total_seconds()
+            if unit == 'h':
+                return timedelta(hours=value).total_seconds()
+            if unit == 'd':
+                return timedelta(days=value).total_seconds()
```
hasegeli: Regexp is overkill to parse something so simple.  How about something like this:
```
units = [
    ('s', 1)
    ('min', 60),
    ('h', 60 * 60),
]

for unit, mul in units:
    if self.period.lower().endswith(unit):
        return int(self.period[:len(unit)].strip()) * mul
```

src/logfile_values.py
```diff
@@ -7,42 +7,298 @@
...
+            elif value == '+':
+                result = result + float(fields[int(arr[index + 1])])
+            elif value == '-':
+                result = result - float(fields[int(arr[index + 1])])
+        return result
+
+    def get_duration(self):
+        '''Returns real duration when metric appeared in log file'''
+        duration = self.end_timestamp - self.start_timestamp
+        return duration
```
hasegeli: This function looks unnecessary.

src/logfile_values.py
```diff
@@ -7,42 +7,298 @@
...
+
+    def get_median(self):
+        sorted_list = sorted(self.values)
+        i = len(sorted_list)
+        if not i % 2:
+            return (sorted_list[(i // 2) - 1] + sorted_list[i // 2]) / 2
```
hasegeli: This looks wrong.  `//` would floor the value.  I think you need to use `i // 2` and `(i // 2) + 1`.

src/logfile_values.py
```diff
@@ -7,42 +7,298 @@
...
+                            fields = line.split()
+                            if len(fields) != 5:
+                                continue
+                            fields[0] = convert_to_timestamp(
+                                fields[0], args.time_format)
+                            for metric in args.metric:
+                                if fields[0] > (
+                                        metric.now - metric.get_timeshift()):
+                                    metric.values.append(
+                                        metric.estimate_columns_value(fields))
```
hasegeli: This is still pretty complicated.  Any luck splitting it into a few more functions?

src/logfile_values.py
v0devil: Yeah :-/ Yesterday was looking necessary cuz the old script estimated SiteReqPerSec and QueriesPerSec from this duration in log file but not a real timeframe -_-

src/logfile_values.py
v0devil: +1

src/logfile_values.py
v0devil: I tested with function statistics.median. This variant looks correct

hasegeli: True, I forgot to take into account that the list identifiers start from `0`.

src/logfile_values.py
```diff
@@ -7,42 +7,271 @@
...
+            elif value == '+':
+                result = result + float(fields[int(arr[index + 1])])
+            elif value == '-':
+                result = result - float(fields[int(arr[index + 1])])
+        return result
+
+    def get_median(self):
+        sorted_list = sorted(self.values)
+        i = len(sorted_list)
+        if not i % 2:
```
hasegeli: I just noticed this is wrong.  I don't think `not` should be there.

v0devil: why 
not means the number of elements is odd (reminder == 0), so we try to get one number above and below from center and divide them by 2.
If it is not odd then just taking the middle number

hasegeli: True, all good.  I am sorry for confusion.

PR #85, opened 2018-01-17
Add new module for Unbound

hasegeli:


src/unbound.py
```diff
@@ -0,0 +1,39 @@
...
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='unbound')
+    return parser.parse_args()
+
+
+def parse_unbound_stats():
+    out = Popen(('/usr/sbin/unbound-control', 'stats'), stdout=PIPE).\
+        stdout.read().splitlines()
```
hasegeli: Could you please use parentheses instead of escaping line-ending?

src/unbound.py
```diff
@@ -0,0 +1,39 @@
...
+
+    return {
+        key.replace('total.', ''): val for key, val in (
```
hasegeli: `.replace()` search for occurrences in anywhere in the string.  That is not what you need.

src/unbound.py
```diff
@@ -0,0 +1,39 @@
...
+            stat.split('=') for stat in out
+        ) if key.startswith('total.')
+    }
```
hasegeli: This returns keys under different levels of namespaces.  I don't think it looks good on Graphite.  It would be more useful to replace the dots with underscores.

src/unbound.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+#
+# igcollect - Stats for Unbound DNS cache/resolver
+#
+# Copyright (c) 2016 InnoGames GmbH
```
hasegeli: 2018

src/unbound.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='unbound')
+    return parser.parse_args()
+
+
+def parse_unbound_stats():
+    out = Popen(('/usr/sbin/unbound-control', 'stats'), stdout=PIPE).\
```
hasegeli: This is silently skipping errors.  Why don't you just use `check_output()`?

src/unbound.py
TheFRedFox: Besides two for loops + if + extra methods seems to be a bit too complex to use list comprehension. I suggest to use normal loops or maybe separate the list comprehensions to make it more clear.

igcollect/unbound.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+#
+# igcollect - Stats for Unbound DNS cache/resolver
+#
+# Copyright (c) 2019 InnoGames GmbH
+#
+
+from __future__ import print_function
```
TheFRedFox: Do we still need support for python 2 compatibility?

igcollect/unbound.py
```diff
@@ -0,0 +1,44 @@
...
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='unbound')
+    return parser.parse_args()
+
+
+def parse_unbound_stats():
+    stats = check_output(
+        ['/usr/local/sbin/unbound-control', 'stats'],
```
TheFRedFox: Is the place of the command fixed there or is it possible that it is somewhere else?

igcollect/unbound.py
```diff
@@ -0,0 +1,44 @@
...
+        universal_newlines=True,
+        close_fds=False,
+    ).split()
+
+    out = {}
+    for stat_k, stat_v in (x.split('=') for x in stats):
+        # Report only stats for all threads
+        if stat_k.startswith('total.'):
+            stat_k = stat_k.replace('total.', '').replace('.', '_')
```
TheFRedFox: I assume that you just want to get rid of the 'total.' prefix? Instead of an expensive and maybe also "unsecure" global replacement (I don't know what the content can be), we could also just use the substring:
``` python
stat_k[5:].replace(...)
```

igcollect/unbound.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='unbound')
+    return parser.parse_args()
+
+
+def parse_unbound_stats():
+    stats = check_output(
```
TheFRedFox: Can you add a python comment just above here with an example output of this `unbound-control`, so also a newbie like me knows how it should be parsed then. 😉

igcollect/unbound.py
```diff
@@ -0,0 +1,44 @@
...
+    out = {}
+    for stat_k, stat_v in (x.split('=') for x in stats):
+        # Report only stats for all threads
+        if stat_k.startswith('total.'):
+            stat_k = stat_k.replace('total.', '').replace('.', '_')
+            out[stat_k] = stat_v
+    return out
+
+
+def main():
```
TheFRedFox: I bet you prefer the C-style function declaration order more. But we rather prefer the order of `argparse->main->helper_functions`.

PR #86, opened 2018-02-07
New mysql status

v0devil:
Use mysql.connector + table sizes

### DIFF
--- a/src/mysql.py
+++ b/src/mysql.py
@@ -5,7 +5,11 @@
 # Copyright (c) 2016, InnoGames GmbH
 #
 
-import MySQLdb
+try:
+    from mysql.connector import connect
+except ImportError:
+    from MySQLdb import connect
+
 from argparse import ArgumentParser
 from time import time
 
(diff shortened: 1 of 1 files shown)


src/mysql.py
```diff
@@ -5,25 +5,32 @@
...
 from argparse import ArgumentParser
 from time import time
 
 
 def parse_args():
     parser = ArgumentParser()
     parser.add_argument('--prefix', default='mysql')
+    parser.add_argument('--user', default='root')
+    parser.add_argument('--host', default='localhost')
+    parser.add_argument('--password', default='password')
```
hasegeli: Can we leave it empty as the default?

src/mysql.py
```diff
@@ -52,13 +59,15 @@ def main():
         if row[0] in sysdbs:
             continue
         cur.execute(
-            'SELECT round(DATA_FREE / 1024 / 1024) '
+            'SELECT table_name, round(DATA_FREE / 1024 / 1024), '
+            'round((data_length + index_length), 2) '
             'FROM information_schema.tables '
             'WHERE TABLE_SCHEMA = %s AND DATA_FREE > 0',
```
hasegeli: It is terrible that some columns are written lower case some upper case.  MySQL shows anything under "information_chema" all upper case, but they are strangely case in-sensitive.  I would still write the column names upper case, and the table name as "information_schema.TABLES" in here.

PR #89, opened 2018-03-14
Add python3 support for nginx module

ragged:


src/nginx.py
```diff
@@ -5,10 +5,16 @@
...
 #
 
+import sys
+
 from argparse import ArgumentParser
-from urllib2 import Request, urlopen
 from time import time
 
+if sys.version_info.major == 3:
+    from urllib.request import Request, urlopen
```
kofrezo: Please use try to import and on except use urllib2

kofrezo: See https://github.com/innogames/igvm/pull/14/files

PR #95, opened 2018-04-05
Small fix to enable arch mode 

v0devil:
need parse the newest archive file if data in the main log file is not complete

### DIFF
--- a/src/logfile_values.py
+++ b/src/logfile_values.py
@@ -270,9 +270,9 @@ def main():  # NOQA: C901
     # If the main file was read completely and there is arch flag then
     # check archive files for the presence of a timestamp satisfying condition
     if args.arch and file_was_readed:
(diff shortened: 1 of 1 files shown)


src/logfile_values.py
```diff
@@ -270,9 +270,9 @@ def main():  # NOQA: C901
     # If the main file was read completely and there is arch flag then
     # check archive files for the presence of a timestamp satisfying condition
     if args.arch and file_was_readed:
-        dir_path = os.path.dirname(os.path.realpath(args.file))
-        archive_pattern = re.compile(r'{}\.\d+?\.gz'.format(args.file))
-        for root, dirs, files in os.walk(dir_path):
+        archive_pattern = re.compile(r'{}.1.gz'.format(
```
hasegeli: Shouldn't the dots be escaped?

PR #96, opened 2018-04-13
jmx.py: Add authentication support

seqizz:


src/jmx.py
```diff
@@ -14,6 +14,8 @@
 def parse_args():
     parser = ArgumentParser()
     parser.add_argument('--prefix', default='jmx')
+    parser.add_argument('--username', default='')
+    parser.add_argument('--password', default='')
```
hasegeli: You do not need the `default`s.  They would be none by default.  `--` means the argument is optional.

hasegeli: There are 6 scripts using with `--user`.  There are 2 scripts with `--pass`, 3 with `--password`.  I would certainly name the former as `--user`, and consider renaming the later.

src/jmx.py
```diff
@@ -26,6 +28,20 @@ def main():
...
 
+    if args.username and args.password:
+        auth_list = [
+                    '--user',
+                    args.username,
+                    '--pass',
+                    args.password
+                    ]
+    elif ((not args.username and args.password)
+            or (args.username and not args.password)):
```
hasegeli: `elif bool(args.username) != bool(args.password)`

src/jmx.py
seqizz: I didn't get the last comment. This is just for jmx plugin, so searching `plugin => 'jmx',` usage in puppet manifests didn't give me something else then godkings/warlords couple.
Maybe I am missing something but, are you recommending renaming `username-password` to  `user-pass` here?

PR #97, opened 2018-04-19
TECH-18936 Net filtering

Felixoid:
Add filtering for sending metrics
Add tests for linux_network


Felixoid:
Amount of sending metrics was decreased for all devices except physical.

src/__init__.py
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env python
+
+pass
```
hasegeli: You don't need any content on this file.

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
 from argparse import ArgumentParser
 from time import time
+import os
+
+
+class InterfaceStatistics(object):
+    '''
+        Supported types of interfaces for a metrics sending
+        If there is more than one conditions they multipied by AND
+    '''
```
hasegeli: PEP8 says to use 3 double quotes on docstrings and the content should not be indented.

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+        'bridge': {'bridge': '_check_dir'},
+        'bridge_slave': {'brport': '_check_dir'},
+        'general_slave': {'master': '_check_symlink'},
+        'lo': {("772", ): "_check_type"},
+        'ovs-br0': {'br0': '_check_name'},
+        'ovs-system': {'ovs-system': '_check_name'},
+        'phys': {'device': '_check_symlink'},
+        'tunnel': {("768", "776"): '_check_type'},
+        'vlan': {"DEVTYPE=vlan": '_check_uevent'},
+    }
```
hasegeli: The inner type would better be a tuple like:
```python
        'bond': ('bonding': '_check_dir'), 
```

hasegeli: Functions are first class citizens on Python.  You can just refer to them like:
```python
        'bond': ('bonding': InterfaceStatistics._check_dir),
```

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+
+    def _check_name(self, dev, name):
+        return dev == name
+
+    def _check_symlink(self, dev, symlink):
+        return os.path.islink(os.path.join(self._scn, dev, symlink))
+
+    def _check_type(self, dev, types):
+        _is_tunnel = False
+        with open(os.path.join(self._scn, dev, 'type'), 'r') as ft:
```
hasegeli: `'r'` is already the default.  We just omit it everywhere.

hasegeli: I don't know what `ft` stands for.  Maybe you mean file descriptor, so `fd`.

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+            dev_type = ft.readline().strip()
+            for t in types:
+                if dev_type == t:
+                    _is_tunnel = True
+        return _is_tunnel
+
+    def _check_uevent(self, dev, string):
+        with open(os.path.join(self._scn, dev, 'uevent'), 'r') as ue:
```
hasegeli: `ue`?

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+    def _check_symlink(self, dev, symlink):
+        return os.path.islink(os.path.join(self._scn, dev, symlink))
+
+    def _check_type(self, dev, types):
+        _is_tunnel = False
+        with open(os.path.join(self._scn, dev, 'type'), 'r') as ft:
+            dev_type = ft.readline().strip()
+            for t in types:
+                if dev_type == t:
+                    _is_tunnel = True
```
hasegeli: You can write it like
```python
_is_tunnel = any(dev_type == t for t in types)
```

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+        return os.path.isdir(os.path.join(self._scn, dev, directory))
+
+    def _check_name(self, dev, name):
+        return dev == name
+
+    def _check_symlink(self, dev, symlink):
+        return os.path.islink(os.path.join(self._scn, dev, symlink))
+
+    def _check_type(self, dev, types):
+        _is_tunnel = False
```
hasegeli: There is no need to prefix variables in Python.  They are always in local scope.

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+                # We send all metrics only for physical devices
+                for m in metric_names:
+                    self.netdev_stat[dev][m] = 0
+                    for param in metric_names[m]:
+                        self.netdev_stat[dev][m] += self._read_stat(dev, param)
+            else:
+                for m in virt_metrics:
+                    self.netdev_stat[dev][m] = 0
+                    for param in metric_names[m]:
+                        self.netdev_stat[dev][m] += self._read_stat(dev, param)
```
hasegeli: `if` and `else` blocks look the same.

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+                    prefix, dev.replace('.', '_'), metric,
+                    self.netdev_stat[dev][metric], self.timestamp
+                ))
 
 
 def parse_args():
     parser = ArgumentParser()
-    parser.add_argument('--prefix', default='network')
+    parser.add_argument('-p', '--prefix', default='network')
+    parser.add_argument('-e', '--enabled-types', action='append',
```
hasegeli: Why add those short names?

tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env python
+
+pass
```
hasegeli: This too

tests/test_linux_network.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class TestInterfaceStatistics(unittest.TestCase):
+    _tests_dir = dirname(getfile(currentframe()))
+    _scn_mock = join(_tests_dir, 'linux_network/sys/class/net')
+    _temp_dir = join(_tests_dir, 'linux_network')
+
+    @classmethod
+    def setUpClass(self):
+        import tarfile
```
hasegeli: Why don't you import it on top.

src/linux_network.py
```diff
@@ -2,77 +2,153 @@
...
+                'tx_window_errors',
+                'tx_heartbeat_errors'
+            ],
+        }
+        virt_metrics = [
+            'bytesIn',
+            'bytesOut',
+            'pktsIn',
+            'pktsOut',
+        ]
```
hasegeli: I would move those out.

src/linux_network.py
Felixoid: I'll change all `with open` to `fd` to be consistent

src/linux_network.py
Felixoid: the same as ft

src/linux_network.py
hasegeli: You should order things accordingly.

Does work:

```python
a = 1
b = a
```

Doesn't work:
```python
b = a
a = 1
```

src/linux_network.py
```diff
@@ -2,77 +2,147 @@
...
+    def _check_name(self, dev, name):
+        return dev == name
+
+    def _check_symlink(self, dev, symlink):
+        return os.path.islink(os.path.join(self._scn, dev, symlink))
+
+    def _check_type(self, dev, types):
+        with open(os.path.join(self._scn, dev, 'type'), 'r') as fd:
+            dev_type = fd.readline().strip()
+            return any(dev_type == t for t in types)
```
hasegeli: This like should be outside of `with:`.

Felixoid: done

Felixoid:
Approve from @TuxPowered42 gained

PR #101, opened 2018-05-12
Remove old Postgres scripts

hasegeli:


Makefile
```diff
@@ -10,10 +10,8 @@ all:
 
 install: test
 	mkdir -p ${targetdir}/${libexecdir}/igcollect
-	mkdir -p ${targetdir}/${libexecdir}/igcollect/libigcollect
 	mkdir -p ${targetdir}/share/java
 	install src/*.py		${targetdir}/${libexecdir}/igcollect
-	install src/libigcollect/*.py	${targetdir}/${libexecdir}/igcollect/libigcollect
```
TheFRedFox: Do we maybe want to keep that folder as example where lib scripts should go to?

PR #107, opened 2018-06-04
Provide the correct redis port from config

Woellchen:


.gitignore
```diff
@@ -61,3 +61,6 @@ target/
 
 #Ipython Notebook
 .ipynb_checkpoints
+
+# IDE
+.idea
```
TheFRedFox: Rather use a local ignore file for this. This is not necessary be ignored for the repository.

There are 3 types to ignore files:

1. Push .gitignore file in repo (globally for the whole repository for all developers)
2. Add a .git/info/exclude file inside your local repository (just locally for just this repository)
3. Use global git config (ignore files globally on your local machine for all repositories)

igcollect/redis.py
```diff
@@ -41,16 +44,19 @@ def main():
...
+        content = fd.read().splitlines()
 
-    for line in secret_cfg:
-        line = line.strip()
-        if line.startswith("requirepass"):
-            return line.split(" ")[1].strip()
-    return ''
+    cfg = {}
+    for line in content:
+        parts = line.strip().split(' ')
```
TheFRedFox: Is it really correct with `split(' ')` rather than just `split()`. `split()` already split on all whitespaces, `split(' ')` just splits on one whitespace. Maybe it is really correct like that, but wondering if it is really meant to be or just accidentally called like that.

.gitignore
TheFRedFox: Have a look here: https://help.github.com/articles/ignoring-files/

igcollect/redis.py
TheFRedFox: With `split()` the former `strip()` might also be not needed anymore.

PR #109, opened 2018-06-07
Only authenticate if a password is configured

meikomeik:


igcollect/redis.py
```diff
@@ -18,9 +18,14 @@ def parse_args():
...
-    redis_info = check_output([
-        'redis-cli', '-a', cfg['requirepass'], '-p', cfg['port'], 'info',
-    ])
+
+    cli_command = ['redis-cli']
+    if 'requirepass' in cfg:
+        cli_command.extend(['-a', cfg['requirepass']])
+    if 'port' in cfg:
+        cli_command.extend(['-p', cfg['port']])
+    cli_command.append('info')
```
TheFRedFox: Would be nicer to have it in an own method.  But not a requirement for being merged.

PR #110, opened 2018-06-08
Update jmxcollect

juergenthomann:
Whitespace and dots in thread prefixes will now be replaced with an underscore


seqizz:
New one seems way better, I don't have privileges to approve the merge but tested and seems ok:
image

PR #123, opened 2019-05-13
Make monitoring plugins Python3 compatible

lamaral:
igcollect is using this plugin on the monitoring servers. 
As we deprecate Python2, I made them compatible with Python3 also.

### DIFF
--- a/igcollect/artfiles.py
+++ b/igcollect/artfiles.py
@@ -1,12 +1,21 @@
 #!/usr/bin/env python
 """igcollect - Artfiles Hosting Metrics
 
-Copyright (c) 2016 InnoGames GmbH
+Copyright (c) 2019 InnoGames GmbH
 """
 
+import base64
 from argparse import ArgumentParser
 from time import time
-from urllib import urlopen
+
+try:
+    # Try importing the Python3 packages
+    from urllib.request import Request, urlopen
+    from urllib.parse import urlencode
+except ImportError:
+    # On failure, import the Python2
+    from urllib2 import Request, urlopen
+    from urllib import urlencode
 
 
 def parse_args():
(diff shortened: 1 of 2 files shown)


igcollect/artfiles.py
```diff
@@ -1,12 +1,18 @@
...
-Copyright (c) 2016 InnoGames GmbH
+Copyright (c) 2019 InnoGames GmbH
 """
 
+import base64
 from argparse import ArgumentParser
 from time import time
-from urllib import urlopen
+
+try:
```
TheFRedFox: Please add a comment which of these modules are for Python2 and which are for Python3.

igcollect/artfiles.py
```diff
@@ -1,12 +1,18 @@
-#!/usr/bin/env python
```
TheFRedFox: Why did you remove this line? PEP8 says keep python if it is python2 and python3.

igcollect/artfiles.py
```diff
@@ -31,21 +37,23 @@ def parse_args():
...
     args = parse_args()
-    artfiles_url = (
-        'https://{}:{}@dcp.c.artfiles.de/api/stats/get_estats.html'
-        .format(args.http_user, args.http_pass)
-    )
-    data = urlopen(artfiles_url)
+    request = Request('https://dcp.c.artfiles.de/api/stats/get_estats.html')
+    credentials = '{}:{}'.format(args.http_user, args.http_pass)
+    base64credentials = base64.b64encode(credentials.encode())
+    request.add_header('Authorization',
```
TheFRedFox: Why aren't you using the `auth` parameter of the constructor of Request?
https://github.com/kennethreitz/requests/blob/master/requests/models.py#L214

TheFRedFox: Ah, nevermind, you are using the urllib module not the request module. But then, one could consider to use the request module. ^^

igcollect/artfiles.py
```diff
@@ -31,21 +37,23 @@ def parse_args():
...
+                       'Basic {}'.format(base64credentials.decode()))
+    data = urlopen(request)
     template = args.prefix + '.{}.{}.{}.{} {} ' + str(int(time()))
```
TheFRedFox: Would prefer to have named template entries, though not in PR scope. So, feel free to do it, but not a must have.

igcollect/artfiles.py
lamaral: I used urllib to avoid having external dependencies.

TheFRedFox: Quite sure, that we kinda habe the request module in every place, as it is so common. But you are right, that we do not need it. Keep it like that. Good to go like that.

igcollect/artfiles.py
lamaral: I removed it because the script is executed by explicitly calling the Python binary. `/usr/bin/python3 artfiles.py`.

But looking at PEP8, it should be there in the main file, so I'll add it back.

igcollect/artfiles.py
lamaral: Fixed 😃

igcollect/fastly.py
```diff
@@ -114,7 +115,7 @@ def main(args):
         regions,
         grequests.map((get_service_data_request(api_key, service, {
             'from': start_time, 'to': end_time,
-            'by': interval, 'region': region,
+            'by':   interval, 'region': region,
```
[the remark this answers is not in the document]
lamaral: Editor's fault. It's now gone.

PR #125, opened 2019-07-02
Allow floats for percentage calculations

raimund-s:
We would like to support also log formats wich log ms instead of seconds. Therefore the percentage calculation needs to be able to parse floats.

Currently we support in only:
TimeSite200:2:count_200_percentage:1min
 
With this patch also fractions are supported:
TimeSite200:2:count_0.200_percentage:1min

### DIFF
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -132,9 +132,9 @@ class Metric:
                     if 'percentage' in self.function:
                         return float(
                             self.get_count_percentage(
-                                int(self.function.split('_')[1])))
+                                float(self.function.split('_')[1])))
                     return float(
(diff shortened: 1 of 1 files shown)


TheFRedFox:
As you are touching it anyway, can you please change the method to something like that:
``` python
def get_metric_value(self):
    if not self.values:
        return 0

    if not self.function:
        return float(self.get_last_value())

    if 'count_' not in self.function:
        return float(getattr(self, 'get_' + self.function)())
    elif 'percentage' in self.function:
        return float(
            self.get_count_percentage(
                float(self.function.split('_')[1]))
            )
    else:
        return float(
            self.get_count(float(self.function.split('_')[1]))
        )
```

raimund-s:
seems outdated

PR #126, opened 2019-07-11
logfile_values: Make time column position configurable and add function to calculate values distribution

v0devil:
Needed for parsing nginx access.log to get the distribution of response codes:
```
/usr/bin/python3 -I ./logfile_values.py --prefix=servers.en0w1_twx.software.gentime --file "/var/log/nginx/access.log" --metric "code:8:distribution:1min" --time-column 3 --time-format "[%d/%b/%Y:%H:%M:%S" --arch
servers.en0w1_twx.software.gentime.code.200.0 15 1562854544
servers.en0w1_twx.software.gentime.code.101.0 40 1562854544
servers.en0w1_twx.software.gentime.code.302.0 1 1562854544
```

### DIFF
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -157,7 +157,7 @@ def parse_args():
     parser = ArgumentParser()
     parser.add_argument('--prefix', default='logfile_values')
     parser.add_argument('--file', default='/var/log/messages')
-    parser.add_argument('--columns-num', default='5', type=int)
+    parser.add_argument('--columns-num', default='0', type=int)
     parser.add_argument('--metric', type=Metric, nargs='+')
     parser.add_argument('--time-column', default='0', type=int)
     parser.add_argument('--time-format', default='%Y-%m-%dT%H:%M:%S%z')
@@ -168,7 +168,7 @@ def parse_args():
 
 def get_metrics_values(line, metrics, time_format, columns_num, time_column):
     fields = line.split()
-    if len(fields) != columns_num:
+    if columns_num and len(fields) != columns_num:
         return True
 
     timestamp = convert_to_timestamp(fields[time_column], time_format)
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -142,7 +142,7 @@ class Metric:
         if self.function == 'distribution':
             return getattr(self, 'get_' + self.function)()
 
-        if self.function.startswith('count'):
+        if self.function.startswith('count_'):
             if self.function.endswith('percentage'):
                 return float(
                     self.get_count_percentage(
@@ -168,14 +168,16 @@ def parse_args():
 
 def get_metrics_values(line, metrics, time_format, columns_num, time_column):
     fields = line.split()
-    if len(fields) == columns_num:
-        timestamp = convert_to_timestamp(fields[time_column], time_format)
-        for metric in metrics:
-            if timestamp > metric.now - metric.get_timeshift():
-                value = metric.estimate_columns_value(fields)
-                metric.values.append(value)
-            else:
-                return False
+    if len(fields) != columns_num:
+        return True
+
+    timestamp = convert_to_timestamp(fields[time_column], time_format)
+    for metric in metrics:
+        if timestamp > metric.now - metric.get_timeshift():
+            value = metric.estimate_columns_value(fields)
+            metric.values.append(value)
+        else:
+            return False
     return True
 
 
@@ -226,9 +228,9 @@ def read_logfile_reverse(filename,
                     lines[-1] += segment
                 else:
                     yield get_metrics_values(
-                                            segment, metrics, time_format,
-                                            columns_num, time_column
-                                        )
+                        segment, metrics, time_format,
+                        columns_num, time_column
+                    )
             segment = lines[0]
             for index in range(len(lines) - 1, 0, -1):
                 global_index += 1
@@ -294,12 +296,12 @@ def main():  # NOQA: C901
                     with gzip.open(archive_file, 'rt', encoding='utf-8') as fh:
                         for line in fh:
                             get_metrics_values(
-                                            line,
-                                            args.metric,
-                                            args.time_format,
-                                            args.columns_num,
-                                            args.time_column,
-                                        )
+                                line,
+                                args.metric,
+                                args.time_format,
+                                args.columns_num,
+                                args.time_column,
+                            )
 
     for metric in args.metric:
         template = args.prefix + '.{} {} ' + str(metric.now)
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -125,19 +125,33 @@ class Metric:
         # Speed :-/?
         return self.get_sum() / self.get_timeshift()
 
+    def get_distribution(self):
+        d = {}
+        uniq_values = set(self.values)
+        for v in uniq_values:
+            d[int(v)] = self.values.count(v)
+        return d
+
     def get_metric_value(self):
-        if self.values:
-            if self.function:
-                if 'count_' in self.function:
-                    if 'percentage' in self.function:
-                        return float(
-                            self.get_count_percentage(
-                                int(self.function.split('_')[1])))
-                    return float(
-                        self.get_count(int(self.function.split('_')[1])))
-                return float(getattr(self, 'get_' + self.function)())
+
+        if not self.values:
+            return 0
+
+        if not self.function:
             return float(self.get_last_value())
-        return 0
+
+        if 'distribution' in self.function:
+            return getattr(self, 'get_' + self.function)()
+
+        if 'count_' in self.function:
+            if 'percentage' in self.function:
+                return float(
+                    self.get_count_percentage(
+                        int(self.function.split('_')[1])))
+            return float(
+                self.get_count(int(self.function.split('_')[1])))
+
+        return float(getattr(self, 'get_' + self.function)())
 
 
 def parse_args():
@@ -146,22 +160,22 @@ def parse_args():
     parser.add_argument('--file', default='/var/log/messages')
     parser.add_argument('--columns-num', default='5', type=int)
     parser.add_argument('--metric', type=Metric, nargs='+')
+    parser.add_argument('--time-column', default='0', type=int)
     parser.add_argument('--time-format', default='%Y-%m-%dT%H:%M:%S%z')
     parser.add_argument('--arch', action='store_true')
     parser.add_argument('--debug', '-d', action='store_true')
     return parser.parse_args()
 
 
-def get_metrics_values(line, metrics, time_format, columns_num):
+def get_metrics_values(line, metrics, time_format, columns_num, time_column):
     fields = line.split()
-    if len(fields) == columns_num:
-        timestamp = convert_to_timestamp(fields[0], time_format)
-        for metric in metrics:
-            if timestamp > metric.now - metric.get_timeshift():
-                value = metric.estimate_columns_value(fields)
-                metric.values.append(value)
-            else:
-                return False
+    timestamp = convert_to_timestamp(fields[time_column], time_format)
+    for metric in metrics:
+        if timestamp > metric.now - metric.get_timeshift():
+            value = metric.estimate_columns_value(fields)
+            metric.values.append(value)
+        else:
+            return False
     return True
 
 
@@ -173,6 +187,7 @@ def get_metrics_last_value(line, metrics):
 
 def read_logfile_reverse(filename,
                          columns_num,
+                         time_column,
                          time_format,
                          metrics,
                          buf_size=8192):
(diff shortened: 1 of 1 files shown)


igcollect/logfile_values.py
```diff
@@ -276,12 +293,17 @@ def main():  # NOQA: C901
                         for line in fh:
                             get_metrics_values(line, args.metric,
                                                args.time_format,
-                                               args.columns_num)
+                                               args.columns_num,
+                                               args.time_column,)
```
TheFRedFox: Apply Multiline style code. Example:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

igcollect/logfile_values.py
```diff
@@ -125,19 +125,33 @@ def get_speed(self):
...
-            if self.function:
-                if 'count_' in self.function:
-                    if 'percentage' in self.function:
-                        return float(
-                            self.get_count_percentage(
-                                int(self.function.split('_')[1])))
-                    return float(
-                        self.get_count(int(self.function.split('_')[1])))
-                return float(getattr(self, 'get_' + self.function)())
+
```
TheFRedFox: Unnecessary blank link.

igcollect/logfile_values.py
```diff
@@ -125,19 +125,33 @@ def get_speed(self):
...
+        if not self.values:
+            return 0
+
+        if not self.function:
             return float(self.get_last_value())
-        return 0
+
+        if 'distribution' in self.function:
```
TheFRedFox: Is `in` really correct, or should it be rather `==`? Right now, this would also be triggered when `self.function` is something like `someother_distribution`.

igcollect/logfile_values.py
```diff
@@ -125,19 +125,33 @@ def get_speed(self):
...
+            return getattr(self, 'get_' + self.function)()
+
+        if 'count_' in self.function:
```
TheFRedFox: Is `in` really correct, or should it be rather `self.function.startswith('count_')`? Right now, this would also be triggered when `self.function` is something like `someother_count_something`.

igcollect/logfile_values.py
```diff
@@ -125,19 +125,33 @@ def get_speed(self):
...
+            if 'percentage' in self.function:
```
TheFRedFox: Is `in` really correct, or should it be rather `self.function.endswith('percentage')` to actually match with `count_percentage`? Is it intended to be that unrestrictive?

igcollect/logfile_values.py
```diff
@@ -146,22 +160,22 @@ def parse_args():
...
     fields = line.split()
-    if len(fields) == columns_num:
-        timestamp = convert_to_timestamp(fields[0], time_format)
-        for metric in metrics:
-            if timestamp > metric.now - metric.get_timeshift():
-                value = metric.estimate_columns_value(fields)
-                metric.values.append(value)
-            else:
-                return False
+    timestamp = convert_to_timestamp(fields[time_column], time_format)
```
TheFRedFox: Real question: Is it safe to have no check for a possible KeyError beforehand here?

igcollect/logfile_values.py
```diff
@@ -146,22 +160,22 @@ def parse_args():
...
     parser.add_argument('--time-format', default='%Y-%m-%dT%H:%M:%S%z')
     parser.add_argument('--arch', action='store_true')
     parser.add_argument('--debug', '-d', action='store_true')
     return parser.parse_args()
 
 
-def get_metrics_values(line, metrics, time_format, columns_num):
+def get_metrics_values(line, metrics, time_format, columns_num, time_column):
     fields = line.split()
-    if len(fields) == columns_num:
```
TheFRedFox: Why is it ok to remove this check?

igcollect/logfile_values.py
```diff
@@ -211,18 +226,18 @@ def read_logfile_reverse(filename,
                     lines[-1] += segment
                 else:
                     yield get_metrics_values(segment, metrics, time_format,
```
TheFRedFox: Apply Multiline style code. Example:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

igcollect/logfile_values.py
```diff
@@ -211,18 +226,18 @@ def read_logfile_reverse(filename,
...
-                                             columns_num)
+                                             columns_num, time_column)
             segment = lines[0]
             for index in range(len(lines) - 1, 0, -1):
                 global_index += 1
                 if lines[index]:
                     if global_index == 1:
                         get_metrics_last_value(lines[index], metrics)
                     yield get_metrics_values(lines[index], metrics,
```
TheFRedFox: Apply Multiline style code. Example:
``` python
func(
    argument_1, argument_2,
    'some small string format'.format(...),
    'some bigger bigger bigger string format'
    .format(...)
    even_more_arguments, ...
)
```

igcollect/logfile_values.py
```diff
@@ -276,12 +293,17 @@ def main():  # NOQA: C901
                         for line in fh:
```
TheFRedFox: I know it is not your code and might be out of scope .. but, we should get rid of that indention mess here.

igcollect/logfile_values.py
```diff
@@ -276,12 +293,17 @@ def main():  # NOQA: C901
...
-                                               args.columns_num)
+                                               args.columns_num,
+                                               args.time_column,)
 
     for metric in args.metric:
         template = args.prefix + '.{} {} ' + str(metric.now)
-        print(template.format(metric.name, metric.get_metric_value()))
-
+        value = metric.get_metric_value()
+        if isinstance(value, dict):
```
TheFRedFox: Where and how is it possible that a value now can be a `dict`?

igcollect/logfile_values.py
```diff
@@ -276,12 +293,17 @@ def main():  # NOQA: C901
...
+            for k, v in value.items():
```
TheFRedFox: A `dict` here can not have further `dict`s inside? Is it sure that there is no recursion?

igcollect/logfile_values.py
```diff
@@ -125,19 +125,33 @@ def get_speed(self):
         # Speed :-/?
         return self.get_sum() / self.get_timeshift()
 
+    def get_distribution(self):
+        d = {}
+        uniq_values = set(self.values)
```
TheFRedFox: Do I get it right, that the list `self.values` doesn't hold complex types like other `list`s oder `dict`s?

igcollect/logfile_values.py
v0devil: it`s my code btw))

igcollect/logfile_values.py
v0devil: in 128

igcollect/logfile_values.py
v0devil: self.values are numbers

igcollect/logfile_values.py
v0devil: No it`s not :-{

igcollect/logfile_values.py
```diff
@@ -291,10 +293,13 @@ def main():  # NOQA: C901
...
-                                               args.time_format,
-                                               args.columns_num,
-                                               args.time_column,)
+                            get_metrics_values(
+                                            line,
+                                            args.metric,
+                                            args.time_format,
+                                            args.columns_num,
+                                            args.time_column,
+                                        )
```
TheFRedFox: The indention of the brace is the same as the `get_metrics_values`.

igcollect/logfile_values.py
```diff
@@ -291,10 +293,13 @@ def main():  # NOQA: C901
                         'Parsing archive file: {}'.format(f))
                     with gzip.open(archive_file, 'rt', encoding='utf-8') as fh:
                         for line in fh:
-                            get_metrics_values(line, args.metric,
-                                               args.time_format,
-                                               args.columns_num,
-                                               args.time_column,)
+                            get_metrics_values(
+                                            line,
```
TheFRedFox: Indention of parameters is one more than of `get_metrics_values`.

igcollect/logfile_values.py
```diff
@@ -225,8 +225,10 @@ def read_logfile_reverse(filename,
                 if buffer[-1] is not '\n':
                     lines[-1] += segment
                 else:
-                    yield get_metrics_values(segment, metrics, time_format,
-                                             columns_num, time_column)
+                    yield get_metrics_values(
+                                            segment, metrics, time_format,
+                                            columns_num, time_column
+                                        )
```
TheFRedFox: The indention of the brace is the same as the `yield`.

igcollect/logfile_values.py
```diff
@@ -225,8 +225,10 @@ def read_logfile_reverse(filename,
                 if buffer[-1] is not '\n':
                     lines[-1] += segment
                 else:
-                    yield get_metrics_values(segment, metrics, time_format,
-                                             columns_num, time_column)
+                    yield get_metrics_values(
+                                            segment, metrics, time_format,
```
TheFRedFox: Indention of parameters is one more than of `yield`.

igcollect/logfile_values.py
```diff
@@ -169,13 +168,14 @@ def parse_args():
...
 def get_metrics_values(line, metrics, time_format, columns_num, time_column):
     fields = line.split()
-    timestamp = convert_to_timestamp(fields[time_column], time_format)
-    for metric in metrics:
-        if timestamp > metric.now - metric.get_timeshift():
-            value = metric.estimate_columns_value(fields)
-            metric.values.append(value)
-        else:
-            return False
+    if len(fields) == columns_num:
```
TheFRedFox: Make a return check out of that:
``` python
if len(fields) != columns_num:
    return True

timestamp = ...
...
```

igcollect/logfile_values.py
```diff
@@ -169,13 +168,14 @@ def parse_args():
 
 def get_metrics_values(line, metrics, time_format, columns_num, time_column):
```
TheFRedFox: Semantically the method name is misleading, should rather be `fetch` or `insert` or the like, as it doesn't return the values but insert them into `metrics` parameter.

igcollect/logfile_values.py
```diff
@@ -133,18 +133,17 @@ def get_distribution(self):
...
         if not self.function:
             return float(self.get_last_value())
 
-        if 'distribution' in self.function:
+        if self.function == 'distribution':
             return getattr(self, 'get_' + self.function)()
 
-        if 'count_' in self.function:
-            if 'percentage' in self.function:
+        if self.function.startswith('count'):
```
TheFRedFox: Why just `count` here all of a sudden and not `count_`? That will also now catch functions like `catchy_phrase`.

igcollect/logfile_values.py
```diff
@@ -133,18 +133,17 @@ def get_distribution(self):
...
+            if self.function.endswith('percentage'):
```
TheFRedFox: Actually this also not completely right, as now it would also catch something like `count_something_percentage`, maybe rather compare it with equal direclty, so `self.function == 'count_percentage'.

TheFRedFox: Btw, how like is it that more possible functions will be added later? If so, we should think of a better structure, like a `dict` with function name as key and function pointer as value.

igcollect/logfile_values.py
TheFRedFox: Then maybe you can tackle this thing during the next PR or when you have some time do some refactoring for it. 😉

igcollect/logfile_values.py
v0devil: Actually it is count_200_percentage and so on and, actually, it could stay with "in" there is no sence in all those restrictions if we not properly raising errors. Nothing will happen if someone (only question why) will want to set ololocount instead of count.

igcollect/logfile_values.py
v0devil: And? What is a reason to put "catchy" phrase

TheFRedFox: Sry, typo, actually it should have been `countable_phrase`. It just does not make sense to check if it is a `count_*` function by just checking `count` as prefix. Just add the `_` there again.

igcollect/logfile_values.py
TheFRedFox: That's a good idea, let's do this. Although not in this PR as it is out of scope. But the arguments and function-checking really could need some redesign in general.

igcollect/logfile_values.py
TheFRedFox: Resolve as it is out of scope for this PR.

igcollect/logfile_values.py
v0devil: Did not get

igcollect/logfile_values.py
v0devil: The script just will not work then

PR #129, opened 2020-01-06
feat(url_json): Allow JSON path to extract values

kofrezo:
Add --json-path which takes one or more JSON path expressions and tries
to parse them from given URL result JSON.

* Add --json-path parameter to allow passing JSON path expression
* Add description to help
* Add example usage to help

### DIFF
--- a/igcollect/url_json.py
+++ b/igcollect/url_json.py
@@ -12,13 +12,31 @@ import json
 
 
 def parse_args():
-    parser = ArgumentParser()
+    parser = ArgumentParser(usage=(
+        'python3 igcollect/url_json.py'
+        ' --prefix servers.HOSTNAME.software.uwsgi'
+        ' --url http://localhost:8080'
+        ' --key pid'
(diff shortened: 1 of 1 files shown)


igcollect/url_json.py
```diff
@@ -29,9 +47,20 @@ def main():
 
     template = args.prefix + '.{} {} ' + str(int(time()))
 
-    for key, value in data.items():
-        if key in args.keys or not args.keys:
-            print(template.format(key, value))
+    if args.keys:
```
Woellchen: unfortunately this is a breaking change. before it would print everything if no keys were given. now it would not print anything. to restore the previous behaviour we could have another `if not args.keys and not args.jsonpaths` and print all the data

kofrezo: I see your point. I considered this a bug as it would print just everything but only for first level and it would print broken metrics in case the value is a list or dictionary.

I will make this backwards compatible.

PR #132, opened 2020-03-03
feat(latency_collection):adds ping.py to collect

Jowe1999:
adds functionality to test the stability of our network

### DIFF
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -23,8 +23,11 @@ def parse_args():
         description='Ping collector script for Graphite ')
     parser.add_argument('hosts', nargs='+',
                         help='the hosts to ping')
-    parser.add_argument('-c', '--count', type=int, default=20,
-                        help='how many times the script will try to ping')
+    parser.add_argument('-c', '--count', type=int, default=20, 
+                        choices=range(1,60),
+                        help='how many times the script will try to ping '
+                             '(number has to be bigger '
+                             'than 1 and smaller than 60)')
     parser.add_argument('-p', '--prefix', default='ping',
                         help='the path to the value in Graphite')
     parser.add_argument('-t', '--timeout', type=int, default=300,
@@ -49,29 +52,30 @@ def main():
 def check_pings(prefix, hosts, count, timeout):
     values = pings(hosts, count, timeout)
 
+    timestamp = str(int(time.time()))
+
     for data in values:
-        send(prefix, data)
+        send(prefix, data, timestamp)
     
 
 def pings(hosts, count, timeout):
     hosts_string = " ".join(hosts)
     # -p calculates the time fping waits between single ping probes
-    # fping used 20 probes/min by default und set this value to 1800
+    # with 20 probes and 1800 milliseconds in between it takes roughly 1 minute
     # if you use 30 probes the -p time should goe down, for 10 up
-    # 0 would not make any sense to enter in to this script
     cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
     cmd = cmd.format(count, (20/count) * 1800, timeout, hosts_string)
 
+    # example.com : 105.94 - 104.78
+    # Showing times for each ping. A ping timeout will be shown as '-'.
     output = subprocess.getoutput(cmd).split('\n')
 
     values = []
     for line in output:
         data = {}
-        parts = line.split(':')
-        data['dest'] = parts[0].replace(' ', '').replace('.', '_')
-        pings = parts[1].split(' ')
-        # first value would be empty
-        pings.pop(0)
+        parts = line.split(':', 1)
+        data['dest'] = parts[0].strip(' ').replace('.', '_')
+        pings = parts[1].split()
 
         data['max'] = max(pings)
         data['min'] = min(pings)
@@ -86,9 +90,12 @@ def pings(hosts, count, timeout):
                 total += float(pings[i])
 
         data['pings'] = pings
-
-        avg = total / len(pings)
-        data['avg'] = avg
+        if (len(pings) <= fails):
+            data['avg'] = -1
+        else:
+            avg = total / (len(pings) - fails)
+            data['avg'] = avg
+            
         data['fails'] = fails
 
         values.append(data)
@@ -96,9 +103,9 @@ def pings(hosts, count, timeout):
     return values
 
 
-def send(prefix, data):
+def send(prefix, data, timestamp):
     template = \
-        str(prefix) + '.' + data['dest'] + '.{} {} ' + str(int(time.time()))
+        str(prefix) + '.' + data['dest'] + '.{} {} ' + timestamp
 
     for num in range(len(data['pings'])):
         print(template.format('ping' + str(num + 1), data['pings'][num]))
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -4,10 +4,10 @@
 Copyright (c) 2020 InnoGames GmbH
 """
 
-# this script is used to collect ping data to evaluate availability and latency
-# of our network (this should replace smokeping in the longrun)
+# This script is used to collect ping data to evaluate availability and latency
+# of our network and make it easy to create a Graphite Dashboard
 #
-# Please Not that the package fping has to be installed on your Linux server!
+# Please note that the package fping has to be installed on your Linux server!
 
 
 import argparse
@@ -36,21 +36,26 @@ def parse_args():
 def main():
     args = parse_args()
 
+    chunked_hosts = [args.hosts[i:i + 10]
+                     for i in range(0, len(args.hosts), 10)]
+
+    data = [(args.prefix, chunk, args.count, args.timeout)
+            for chunk in chunked_hosts]
+
     with Pool(cpu_count() * 2) as p:
-        for i in range(len(args.hosts) % 10):
-            p.starmap(ping, [(args.prefix, args.hosts[i*10:i*10+9],
-                              args.count, args.timeout)])
-        p.starmap(ping, [(args.prefix,
-                          args.hosts[(len(args.hosts) % 10) * 10:
-                                     len(args.hosts) - 1],
-                          args.count, args.timeout)])
+        p.starmap(ping, data)
+
 
 
 def ping(prefix, hosts, count, timeout):
     hosts_string = " ".join(hosts)
     # -p calculates the time fping waits between single ping probes
-    cmd = ('fping -B1 -q -C {} -p {} -t {} {}'.format(count, (20/count) * 1800,
-                                                      timeout, hosts_string))
+    # fping used 20 probes/min by default und set this value to 1800
+    # if you use 30 probes the -p time should goe down, for 10 up
+    # 0 would not make any sense to enter in to this script
+    cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
+    cmd = cmd.format(count, (20/count) * 1800, timeout, hosts_string)
+
     output = subprocess.getoutput(cmd).split('\n')
     for line in output:
         data = {}
@@ -60,7 +65,6 @@ def ping(prefix, hosts, count, timeout):
         # first value would be empty
         pings.pop(0)
 
-        print(pings)
         avg = 0
         for i, ping in enumerate(pings):
             if pings[i] == '-':
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -13,6 +13,7 @@ Copyright (c) 2020 InnoGames GmbH
 import argparse
 import subprocess
 import time
+import socket
 
 from multiprocessing import Pool
 from os import cpu_count
@@ -28,7 +29,9 @@ def parse_args():
                         help='how many times the script will try to ping '
                              '(number has to be bigger '
                              'than 1 and smaller than 60) (default= 20)')
-    parser.add_argument('-p', '--prefix', default='ping',
+    parser.add_argument('-p', '--prefix',
+                        default='servers.{}.system.ping'.format(
+                            socket.gethostname()),
                         help='the path to the value in Graphite '
                              '(default= ping)')
     parser.add_argument('-t', '--timeout', type=int, default=300,
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -21,15 +21,15 @@ from os import cpu_count
 def parse_args():
     parser = argparse.ArgumentParser(
         description='Ping collector script for Graphite ')
-    parser.add_argument('hosts', metavar='h', nargs='+',
-                        help='the adresses to ping')
+    parser.add_argument('hosts', nargs='+',
+                        help='the hosts to ping')
     parser.add_argument('-c', '--count', type=int, default=20,
-                        help='how many times should the script tries to ping')
+                        help='how many times the script will try to ping')
     parser.add_argument('-p', '--prefix', default='ping',
                         help='the path to the value in Graphite')
     parser.add_argument('-t', '--timeout', type=int, default=300,
-                        help='time in milliseconds'
-                             ' till the timeout is retched')
+                        help='time in milliseconds '
+                             'till the timeout is retched')
     return parser.parse_args()
 
 
@@ -43,11 +43,17 @@ def main():
             for chunk in chunked_hosts]
 
     with Pool(cpu_count() * 2) as p:
-        p.starmap(ping, data)
+        p.starmap(check_pings, data)
 
 
+def check_pings(prefix, hosts, count, timeout):
+    values = pings(hosts, count, timeout)
 
-def ping(prefix, hosts, count, timeout):
+    for data in values:
+        send(prefix, data)
+    
+
+def pings(hosts, count, timeout):
     hosts_string = " ".join(hosts)
     # -p calculates the time fping waits between single ping probes
     # fping used 20 probes/min by default und set this value to 1800
@@ -57,6 +63,8 @@ def ping(prefix, hosts, count, timeout):
     cmd = cmd.format(count, (20/count) * 1800, timeout, hosts_string)
 
     output = subprocess.getoutput(cmd).split('\n')
+
+    values = []
     for line in output:
         data = {}
         parts = line.split(':')
@@ -65,41 +73,42 @@ def ping(prefix, hosts, count, timeout):
         # first value would be empty
         pings.pop(0)
 
-        avg = 0
-        for i, ping in enumerate(pings):
-            if pings[i] == '-':
-                data['ping{}'.format(i)] = -1
-            else:
-                data['ping{}'.format(i)] = pings[i]
-                avg += float(pings[i])
-            i += 1
-
-        avg /= len(pings)
-        data['avg'] = avg
-
-        pings = pings[0:]
         data['max'] = max(pings)
         data['min'] = min(pings)
+        
+        total = 0
+        fails = 0
+        for i, ping in enumerate(pings):
+            if ping == '-':
+                pings[i] = -1
+                fails += 1
+            else:
+                total += float(pings[i])
 
-        send(prefix, data, count)
+        data['pings'] = pings
+
+        avg = total / len(pings)
+        data['avg'] = avg
+        data['fails'] = fails
+
+        values.append(data)
+
+    return values
 
 
-def send(prefix, data, count):
-    template = str(prefix) + '.{}.{} {} ' + str(int(time.time()))
+def send(prefix, data):
+    template = \
+        str(prefix) + '.' + data['dest'] + '.{} {} ' + str(int(time.time()))
 
-    fails = 0
-    for num in range(count):
-        print(template.format(data['dest'], 'ping' + str(num + 1),
-                              data['ping' + str(num)]))
-        if data['ping' + str(num)] == -1:
-            fails += 1
+    for num in range(len(data['pings'])):
+        print(template.format('ping' + str(num + 1), data['pings'][num]))
 
-    print(template.format(data['dest'], 'fails', fails))
-    print(template.format(data['dest'], 'fails/1', fails / count))
+    print(template.format('fails', data['fails']))
+    print(template.format('fails/1', data['fails'] / len(data['pings'])))
 
-    print(template.format(data['dest'], 'min', data['min']))
-    print(template.format(data['dest'], 'max', data['max']))
-    print(template.format(data['dest'], 'avg', data['avg']))
+    print(template.format('min', data['min']))
+    print(template.format('max', data['max']))
+    print(template.format('avg', data['avg']))
 
 
 if __name__ == '__main__':
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -31,9 +31,9 @@ def parse_args():
                              'than 1 and smaller than 60) (default= 20)')
     parser.add_argument('-p', '--prefix',
                         default='servers.{}.system.ping'.format(
-                            socket.gethostname()),
+                            socket.gethostname().replace('.', '_')),
                         help='the path to the value in Graphite '
-                             '(default= ping)')
+                             '(default= servers.(hostname).system.ping)')
     parser.add_argument('-t', '--timeout', type=int, default=300,
                         help='time in milliseconds '
                              'till the timeout is retched (default= 300)')
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -17,6 +17,7 @@ import socket
 
 from multiprocessing import Pool
 from os import cpu_count
+from statistics import stdev
 
 
 def parse_args():
@@ -24,7 +25,7 @@ def parse_args():
         description='Ping collector script for Graphite ')
     parser.add_argument('hosts', nargs='+',
                         help='the hosts to ping')
-    parser.add_argument('-c', '--count', type=int, default=20, 
+    parser.add_argument('-c', '--count', type=int, default=20,
                         choices=range(1,60),
                         help='how many times the script will try to ping '
                              '(number has to be bigger '
@@ -34,13 +35,14 @@ def parse_args():
                             socket.gethostname().replace('.', '_')),
                         help='the path to the value in Graphite '
                              '(default= servers.(hostname).system.ping)')
-    parser.add_argument('-t', '--timeout', type=int, default=300,
-                        help='time in milliseconds '
-                             'till the timeout is retched (default= 300)')
+    parser.add_argument('-t', '--timeout', type=int, default=500,
+                        help='Initial target timeout in milliseconds. '
+                             'the amount of time that program '
+                             'waits for a response (default= 500)')
     parser.add_argument('-d', '--delay', type=int, default=1800,
                         choices=range(0,5000),
-                        help='time to wait between a series of pings '
-                             '(default= 1800)')
+                        help='time to wait between a series of pings in '
+                             'milliseconds (default= 1800 for 20 pings)')
     return parser.parse_args()
 
 
@@ -64,15 +66,15 @@ def check_pings(prefix, hosts, count, timeout, delay):
 
     for data in values:
         print_ping(prefix, data, timestamp)
-    
+
 
 def pings(hosts, count, timeout, delay):
     hosts_string = ' '.join(hosts)
     # with 20 probes and 1800 milliseconds in between it takes roughly 1 minute
     # the default value for delay is 1800
     cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
-    cmd = cmd.format(count, (20/count) * delay, timeout, hosts_string)
-
+    cmd = cmd.format(count, delay, timeout, hosts_string)
+    # output:
     # example.com : 105.94 - 104.78
     # Showing times for each ping. A ping timeout will be shown as '-'.
     output = subprocess.getoutput(cmd).split('\n')
@@ -84,17 +86,19 @@ def pings(hosts, count, timeout, delay):
         data['dest'] = parts[0].strip(' ').replace('.', '_')
         pings = parts[1].split()
 
-        data['max'] = max(pings)
-        data['min'] = min(pings)
-        
         total = 0.0
         fails = 0
         for i, ping in enumerate(pings):
             if ping == '-':
-                pings[i] = -1
+                pings[i] = '-1'
                 fails += 1
             else:
                 total += float(ping)
+
+        pings = list(map(float, pings))
+
+        data['max'] = max(pings)
+        data['min'] = min(pings)
 
         data['pings'] = pings
         if (len(pings) <= fails):
@@ -102,8 +106,10 @@ def pings(hosts, count, timeout, delay):
         else:
             avg = total / (len(pings) - fails)
             data['avg'] = avg
-            
+
         data['fails'] = fails
+
+        data['std'] = stdev(pings)
 
         values.append(data)
 
@@ -121,10 +127,11 @@ def print_ping(prefix, data, timestamp):
     print(template.format('fails', data['fails']))
 
     if len(data['pings']) == 0:
-        raise Exception('something went horrible wrong - check your fping -')
+        raise Exception('something went horrible wrong: check your fping')
 
     print(template.format('fails/1', data['fails'] / len(data['pings'])))
 
+    print(template.format('std', data['std']))
     print(template.format('min', data['min']))
     print(template.format('max', data['max']))
     print(template.format('avg', data['avg']))
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -70,7 +70,7 @@ def check_pings(prefix, hosts, count, timeout, delay):
     """ checks the pings and prints them
 
         Parameters:
-            prefix (str): were the script will put the data in graphite
+            prefix (str): the prefix for the output for graphite
             hosts (list(str)):  all the hosts that get pinged
             count (int): how many times a individual host get pinged
             timeout (int): the time till the timeout of fping
@@ -148,7 +148,7 @@ def print_ping(prefix, data, timestamp):
 
         Parameters:
             prefix (str): were the script will put the data in graphite
-            data (dict(floar)): the data gathered (pings, min, may, avg, std)
+            data (dict(floar)): the data gathered (pings, min, max, avg, std)
             timestamp (str): the timestamp as a number
 
     """
@@ -164,7 +164,7 @@ def print_ping(prefix, data, timestamp):
 
     print(template.format('fails', data['fails']))
 
-    if len(pings):
+    if pings:
         # If the length is 0 we ignore the entries
 
         print(template.format('fails/1', data['fails'] / len(pings)))
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -27,12 +27,17 @@ def parse_args():
                         choices=range(1,60),
                         help='how many times the script will try to ping '
                              '(number has to be bigger '
-                             'than 1 and smaller than 60)')
+                             'than 1 and smaller than 60) (default= 20)')
     parser.add_argument('-p', '--prefix', default='ping',
-                        help='the path to the value in Graphite')
+                        help='the path to the value in Graphite '
+                             '(default= ping)')
     parser.add_argument('-t', '--timeout', type=int, default=300,
                         help='time in milliseconds '
-                             'till the timeout is retched')
+                             'till the timeout is retched (default= 300)')
+    parser.add_argument('-d', '--delay', type=int, default=1800,
+                        choices=range(0,5000),
+                        help='time to wait between a series of pings '
+                             '(default= 1800)')
     return parser.parse_args()
 
 
@@ -42,29 +47,28 @@ def main():
     chunked_hosts = [args.hosts[i:i + 10]
                      for i in range(0, len(args.hosts), 10)]
 
-    data = [(args.prefix, chunk, args.count, args.timeout)
+    data = [(args.prefix, chunk, args.count, args.timeout, args.delay)
             for chunk in chunked_hosts]
 
     with Pool(cpu_count() * 2) as p:
         p.starmap(check_pings, data)
 
 
-def check_pings(prefix, hosts, count, timeout):
-    values = pings(hosts, count, timeout)
+def check_pings(prefix, hosts, count, timeout, delay):
+    values = pings(hosts, count, timeout, delay)
 
     timestamp = str(int(time.time()))
 
     for data in values:
-        send(prefix, data, timestamp)
+        print_ping(prefix, data, timestamp)
     
 
-def pings(hosts, count, timeout):
-    hosts_string = " ".join(hosts)
-    # -p calculates the time fping waits between single ping probes
+def pings(hosts, count, timeout, delay):
+    hosts_string = ' '.join(hosts)
     # with 20 probes and 1800 milliseconds in between it takes roughly 1 minute
-    # if you use 30 probes the -p time should goe down, for 10 up
+    # the default value for delay is 1800
     cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
-    cmd = cmd.format(count, (20/count) * 1800, timeout, hosts_string)
+    cmd = cmd.format(count, (20/count) * delay, timeout, hosts_string)
 
     # example.com : 105.94 - 104.78
     # Showing times for each ping. A ping timeout will be shown as '-'.
@@ -80,14 +84,14 @@ def pings(hosts, count, timeout):
         data['max'] = max(pings)
         data['min'] = min(pings)
         
-        total = 0
+        total = 0.0
         fails = 0
         for i, ping in enumerate(pings):
             if ping == '-':
                 pings[i] = -1
                 fails += 1
             else:
-                total += float(pings[i])
+                total += float(ping)
 
         data['pings'] = pings
         if (len(pings) <= fails):
@@ -103,14 +107,19 @@ def pings(hosts, count, timeout):
     return values
 
 
-def send(prefix, data, timestamp):
-    template = \
+def print_ping(prefix, data, timestamp):
+    template = (
         str(prefix) + '.' + data['dest'] + '.{} {} ' + timestamp
+    )
 
-    for num in range(len(data['pings'])):
-        print(template.format('ping' + str(num + 1), data['pings'][num]))
+    for num, ping in enumerate(data['pings'], 1):
+        print(template.format('ping' + str(num), ping))
 
     print(template.format('fails', data['fails']))
+
+    if len(data['pings']) == 0:
+        raise Exception('something went horrible wrong - check your fping -')
+
     print(template.format('fails/1', data['fails'] / len(data['pings'])))
 
     print(template.format('min', data['min']))
--- a/igcollect/ping.py
+++ b/igcollect/ping.py
@@ -4,56 +4,65 @@
 Copyright (c) 2020 InnoGames GmbH
 """
 
-import _thread
+# this script is used to collect ping data to evaluate availability and latency
+# of our network (this should replace smokeping in the longrun)
+#
+# Please Not that the package fping has to be installed on your Linux server!
+
+
 import argparse
 import subprocess
+import time
 
-from time import time, sleep
+from multiprocessing import Pool
+from os import cpu_count
+
 
 def parse_args():
     parser = argparse.ArgumentParser(
-        description='Ping colecctor script for Graphana ')
-    parser.add_argument('adress', metavar='URL', type=str, nargs='+',
-                        help='the address zu ping to')
+        description='Ping collector script for Graphite ')
+    parser.add_argument('hosts', metavar='h', nargs='+',
+                        help='the adresses to ping')
     parser.add_argument('-c', '--count', type=int, default=20,
-                        help='how many tims should the script tries to ping')
-    parser.add_argument('--prefix', default='ping',
-                        help='the path to the value in Graphana')
+                        help='how many times should the script tries to ping')
+    parser.add_argument('-p', '--prefix', default='ping',
+                        help='the path to the value in Graphite')
+    parser.add_argument('-t', '--timeout', type=int, default=300,
+                        help='time in milliseconds'
+                             ' till the timeout is retched')
     return parser.parse_args()
 
 
 def main():
     args = parse_args()
-    hostname = args.adress
-    count = args.count
-    prefix = args.prefix
 
-    i = 1
-    hosts = ''
-    for host in hostname:
-        hosts += host + ' '
-        if i % 10 == 0:
-            _thread.start_new_thread(ping, (prefix, hosts, count))
-            hosts = ''
-    if hosts != '':
-        ping(prefix, hosts, count)
-
-    while _thread._count() > 0:
-        sleep(1)
+    with Pool(cpu_count() * 2) as p:
+        for i in range(len(args.hosts) % 10):
+            p.starmap(ping, [(args.prefix, args.hosts[i*10:i*10+9],
+                              args.count, args.timeout)])
+        p.starmap(ping, [(args.prefix,
+                          args.hosts[(len(args.hosts) % 10) * 10:
+                                     len(args.hosts) - 1],
+                          args.count, args.timeout)])
 
 
-def ping(prefix, hosts, count):
-    cmd = ('fping -B1 -q -C {} -p {} {}'.format(count, (20/count)*2000, hosts))
-    fping = subprocess.getoutput(cmd)
-    output = fping.split('\n')
+def ping(prefix, hosts, count, timeout):
+    hosts_string = " ".join(hosts)
+    # -p calculates the time fping waits between single ping probes
+    cmd = ('fping -B1 -q -C {} -p {} -t {} {}'.format(count, (20/count) * 1800,
+                                                      timeout, hosts_string))
+    output = subprocess.getoutput(cmd).split('\n')
     for line in output:
         data = {}
-        part = line.split(':')
-        data['dest'] = part[0].replace(' ', '').replace('.', '_')
-        pings = part[1].split(' ')
-        i = 1
+        parts = line.split(':')
+        data['dest'] = parts[0].replace(' ', '').replace('.', '_')
+        pings = parts[1].split(' ')
+        # first value would be empty
+        pings.pop(0)
+
+        print(pings)
         avg = 0
-        while i <= count:
+        for i, ping in enumerate(pings):
             if pings[i] == '-':
                 data['ping{}'.format(i)] = -1
             else:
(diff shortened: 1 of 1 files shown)


igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
+
+"""igcollect - Latency collection
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+import _thread
```
TuxPowered42: Are you sure you want the low level approach? There are high level libraries for that. Especially with Python3 there is `concurrent` module.

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+            i += 1
+
+        avg /= count
+        if avg == 0.0:
+            data['avg'] = '-'
+        else:
+            data['avg'] = avg
+        pings = pings[1:]
+        data['max'] = max(pings)
+        data['min'] = min(pings)
```
TuxPowered42: Shouldn't Graphite perform min/max/avg calculations from raw data?

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+
+import _thread
+import argparse
+import subprocess
+
+from time import time, sleep
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Ping colecctor script for Graphana ')
```
Woellchen: typo: `Grafana`

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+    parser.add_argument('adress', metavar='URL', type=str, nargs='+',
```
Woellchen: typo: `address`

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+                        help='the address zu ping to')
+    parser.add_argument('-c', '--count', type=int, default=20,
+                        help='how many tims should the script tries to ping')
```
Woellchen: typo: `times` and `try`

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+    parser.add_argument('--prefix', default='ping',
+                        help='the path to the value in Graphana')
```
Woellchen: typo: `Grafana`

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    hostname = args.adress
```
Woellchen: it would be beneficial to keep the same name. decide on either `hostname` or `address`

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+    count = args.count
+    prefix = args.prefix
+
+    i = 1
+    hosts = ''
+    for host in hostname:
```
Woellchen: instead of tracking `i` separately just do `for i, host in enumerate(hostname)`. it will start at zero though. keep that in mind

igcollect/ping.py
Woellchen: also this accepts multiple addresses. so rather name it `addresses`

igcollect/ping.py
TuxPowered42: s/colecctor/collector/
s/Graphana/Grafana/

igcollect/ping.py
TuxPowered42: s/adress/address/

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+import argparse
+import subprocess
+
+from time import time, sleep
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Ping colecctor script for Graphana ')
+    parser.add_argument('adress', metavar='URL', type=str, nargs='+',
+                        help='the address zu ping to')
```
TuxPowered42: s/zu/$I_DONT_EVEN_KNOW_WHAT/

igcollect/ping.py
TuxPowered42: I give up, just fix all spelling by yourself :)

igcollect/ping.py
TuxPowered42: Looks redundant to me.

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+    hostname = args.adress
+    count = args.count
+    prefix = args.prefix
+
+    i = 1
+    hosts = ''
+    for host in hostname:
+        hosts += host + ' '
+        if i % 10 == 0:
+            _thread.start_new_thread(ping, (prefix, hosts, count))
```
Woellchen: why use threading in the first place? and when we are using it, why multiple hosts at once?

Woellchen: next question why use `_thread`? you have already `subprocess` imported. use that one instead

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+            hosts = ''
+    if hosts != '':
+        ping(prefix, hosts, count)
+
+    while _thread._count() > 0:
+        sleep(1)
+
+
+def ping(prefix, hosts, count):
+    cmd = ('fping -B1 -q -C {} -p {} {}'.format(count, (20/count)*2000, hosts))
```
Woellchen: two magic numbers here. what is `20` and what is `2000`? they should be constants at the very least, or even better optional arguments to the script

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+    fping = subprocess.getoutput(cmd)
+    output = fping.split('\n')
+    for line in output:
+        data = {}
+        part = line.split(':')
```
Woellchen: `split` returns an array. so the variable should be named `parts`

igcollect/ping.py
TuxPowered42: Some comment about used numbers would be useful. I can't say from that command line if there is anything that guarantees fping to finish within one minute.

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+        data['dest'] = part[0].replace(' ', '').replace('.', '_')
+        pings = part[1].split(' ')
+        i = 1
+        avg = 0
+        while i <= count:
```
Woellchen: `for i in range(count)` or maybe even better `for i, ping in enumerate(pings)`

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+
+    while _thread._count() > 0:
+        sleep(1)
+
+
+def ping(prefix, hosts, count):
+    cmd = ('fping -B1 -q -C {} -p {} {}'.format(count, (20/count)*2000, hosts))
+    fping = subprocess.getoutput(cmd)
+    output = fping.split('\n')
+    for line in output:
```
TuxPowered42: Lines above could all be in one line.

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+        avg = 0
+        while i <= count:
+            if pings[i] == '-':
+                data['ping{}'.format(i)] = -1
+            else:
+                data['ping{}'.format(i)] = pings[i]
+                avg += float(pings[i])
+            i += 1
+
+        avg /= count
```
Woellchen: `avg /= len(pings)` would be safer

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+        if avg == 0.0:
```
Woellchen: in which case is it possible that `avg` is zero?

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+            data['avg'] = '-'
+        else:
+            data['avg'] = avg
+        pings = pings[1:]
```
Woellchen: why are we discarding the first ping?

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+
+    fails = 0
+    for num in range(count):
+        print(template.format(data['dest'], 'ping' + str(num + 1),
+                              data['ping' + str(num + 1)]))
+        if data['ping' + str(num + 1)] == -1:
+            fails += 1
+
+    print(template.format(data['dest'], 'fails', fails))
+    print(template.format(data['dest'], 'fails/1', fails / count))
```
Woellchen: what is `fails/1`?

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
+
+"""igcollect - Latency collection
```
Woellchen: i would expect some comments up here that describes the purpose of the script and maybe some usage examples

igcollect/ping.py
```diff
@@ -0,0 +1,95 @@
...
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+import _thread
+import argparse
+import subprocess
+
+from time import time, sleep
```
Woellchen: you should rather `import time` and use `time.time()` and `time.sleep()` respectively. otherwise you would shadow the package name `time`

igcollect/ping.py
Woellchen: yes

igcollect/ping.py
Woellchen: scratch that. kajetan is right. graphite should do the aggregations

igcollect/ping.py
Woellchen: scratch that. kajetan is right. graphite should do the aggregations

igcollect/ping.py
axxelG: There are two approaches:
1. Store every ping result
2. Do a "one minute 100s of pings" check and safe the statistical data.
It depends on what you like to know.
I assume we will do this on a lot of servers so my approach would be to do less pings per host but store every result.

igcollect/ping.py
TheFRedFox: `enumerate` also takes a starting number, so it could be: `enumerate(hostname, 1)`
However, `i` is never changed inside the for-loop. Is it really needed?

igcollect/ping.py
TuxPowered42: Furthermore it's hosts that are pinged, not necessarily addresses. I assume fping will query DNS.

igcollect/ping.py
kofrezo: functions and modules starting with underscore are not meant to be used outside. They may disappear, change its behaviour at any time.

igcollect/ping.py
kofrezo: typo: collector

igcollect/ping.py
kofrezo: - string is the default type
- Why name it address and then internally use URL ?

igcollect/ping.py
kofrezo: Provide shortcut for all (address = -a) or none (consistency)

igcollect/ping.py
Jowe1999: changed to multiprosessing and simpyfied

igcollect/ping.py
Jowe1999: the dashbord overloads if i calculate this for to many pings on the board

igcollect/ping.py
Jowe1999: ttl added

igcollect/ping.py
Jowe1999: compleatly changed

igcollect/ping.py
Jowe1999: this calculates the time between individual pings (itst the same than smokeping)

igcollect/ping.py
Jowe1999: parts from one ^^

igcollect/ping.py
Jowe1999: it's empty (caused by the scraping of the output)

igcollect/ping.py
TuxPowered42: In this case please also add standard deviation.

igcollect/ping.py
Jowe1999: thats clever I like ^^

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
-    fping = subprocess.getoutput(cmd)
-    output = fping.split('\n')
+
+    with Pool(cpu_count() * 2) as p:
+        for i in range(len(args.hosts) % 10):
+            p.starmap(ping, [(args.prefix, args.hosts[i*10:i*10+9],
+                              args.count, args.timeout)])
+        p.starmap(ping, [(args.prefix,
+                          args.hosts[(len(args.hosts) % 10) * 10:
+                                     len(args.hosts) - 1],
```
TheFRedFox: Please rethink the calculation of the rest of the hosts. E.g. go through the example of having 9 hosts.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
-        sleep(1)
-
-
-def ping(prefix, hosts, count):
-    cmd = ('fping -B1 -q -C {} -p {} {}'.format(count, (20/count)*2000, hosts))
-    fping = subprocess.getoutput(cmd)
-    output = fping.split('\n')
+
+    with Pool(cpu_count() * 2) as p:
+        for i in range(len(args.hosts) % 10):
```
TheFRedFox: How about chunking up the hosts before this to separate scopes here:
```suggestion
        chunked_hosts = [ x[i:i + 10] for i in range(0, len(args.hosts), 10)]
        for hosts in chunked_hosts:
```

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+        p.starmap(ping, [(args.prefix,
+                          args.hosts[(len(args.hosts) % 10) * 10:
+                                     len(args.hosts) - 1],
+                          args.count, args.timeout)])
+
+
+def ping(prefix, hosts, count, timeout):
+    hosts_string = " ".join(hosts)
+    # -p calculates the time fping waits between single ping probes
+    cmd = ('fping -B1 -q -C {} -p {} -t {} {}'.format(count, (20/count) * 1800,
```
TheFRedFox: How about:
```suggestion
    cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
    cmd = cmd.format(count, (20/count) * 1800, timeout, hosts_string)
```

TheFRedFox: Again, you are not explaining those magic numbers `20` and `1800` for the calculation of the frequency.
Where does specifically `1800` comes from, where `20`?
What will happen if you pass `0` for `count`?

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
 Copyright (c) 2020 InnoGames GmbH
 """
 
-import _thread
+# this script is used to collect ping data to evaluate availability and latency
+# of our network (this should replace smokeping in the longrun)
+#
+# Please Not that the package fping has to be installed on your Linux server!
```
TheFRedFox: How safe is it to rely on `fping` and how important is it? We are already using multiprocessing, how balanced are the pros/cons using `fping` over `ping` here?

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
-from time import time, sleep
 
 def parse_args():
     parser = argparse.ArgumentParser(
-        description='Ping colecctor script for Graphana ')
-    parser.add_argument('adress', metavar='URL', type=str, nargs='+',
-                        help='the address zu ping to')
+        description='Ping collector script for Graphite ')
+    parser.add_argument('hosts', metavar='h', nargs='+',
+                        help='the adresses to ping')
```
TheFRedFox: You are again mixing terms here: `hosts` and `addresses` in the help. Stick to one and `hosts` should be more correct here.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
 
-from time import time, sleep
 
 def parse_args():
     parser = argparse.ArgumentParser(
-        description='Ping colecctor script for Graphana ')
-    parser.add_argument('adress', metavar='URL', type=str, nargs='+',
-                        help='the address zu ping to')
+        description='Ping collector script for Graphite ')
+    parser.add_argument('hosts', metavar='h', nargs='+',
```
TheFRedFox: Rethink again what the purpose of the `metavar` is and if `h` is a good naming.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
     parser.add_argument('-c', '--count', type=int, default=20,
-                        help='how many tims should the script tries to ping')
-    parser.add_argument('--prefix', default='ping',
-                        help='the path to the value in Graphana')
+                        help='how many times should the script tries to ping')
+    parser.add_argument('-p', '--prefix', default='ping',
+                        help='the path to the value in Graphite')
+    parser.add_argument('-t', '--timeout', type=int, default=300,
+                        help='time in milliseconds'
+                             ' till the timeout is retched')
```
TheFRedFox:
```suggestion
                        help='time in milliseconds '
                             'till the timeout is retched')
```

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
-    parser.add_argument('adress', metavar='URL', type=str, nargs='+',
-                        help='the address zu ping to')
+        description='Ping collector script for Graphite ')
+    parser.add_argument('hosts', metavar='h', nargs='+',
+                        help='the adresses to ping')
     parser.add_argument('-c', '--count', type=int, default=20,
-                        help='how many tims should the script tries to ping')
-    parser.add_argument('--prefix', default='ping',
-                        help='the path to the value in Graphana')
+                        help='how many times should the script tries to ping')
```
TheFRedFox: This is a question and not a statement:
```suggestion
                        help='how many times the script will try to ping')
```

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+        for i in range(len(args.hosts) % 10):
+            p.starmap(ping, [(args.prefix, args.hosts[i*10:i*10+9],
+                              args.count, args.timeout)])
+        p.starmap(ping, [(args.prefix,
+                          args.hosts[(len(args.hosts) % 10) * 10:
+                                     len(args.hosts) - 1],
+                          args.count, args.timeout)])
+
+
+def ping(prefix, hosts, count, timeout):
```
TheFRedFox: To pass `prefix` to a utility function like `ping` seems odd. How about splitting this up a bit into:
```
main()
    Pool.map(check_pings)
check_pings():
    pings()
    send()
ping()
send()
```

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+    hosts_string = " ".join(hosts)
+    # -p calculates the time fping waits between single ping probes
+    cmd = ('fping -B1 -q -C {} -p {} -t {} {}'.format(count, (20/count) * 1800,
+                                                      timeout, hosts_string))
+    output = subprocess.getoutput(cmd).split('\n')
```
TheFRedFox: Reviewing is quite hard without knowing the output. Please comment in a short example what the output could be, so it will get easier to follow up the next steps.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+        pings = parts[1].split(' ')
+        # first value would be empty
+        pings.pop(0)
+
+        print(pings)
         avg = 0
-        while i <= count:
+        for i, ping in enumerate(pings):
             if pings[i] == '-':
                 data['ping{}'.format(i)] = -1
```
TheFRedFox: How about not having the structure like:
```
data[
    'dest': 'host'
    'ping1': 1
    'ping2': -1
    ...
]
```
but:
```
dest[
    'dest': 'hosts'
    pings: [1, -1, ...]
]
```

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
             else:
                 data['ping{}'.format(i)] = pings[i]
                 avg += float(pings[i])
```
TheFRedFox: How about counting it up in a `total` variable and then calculate it below with:
``` python
avg = total / len(pings)
```

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
-        avg /= count
-        if avg == 0.0:
-            data['avg'] = '-'
-        else:
-            data['avg'] = avg
-        pings = pings[1:]
+        avg /= len(pings)
+        data['avg'] = avg
+
+        pings = pings[0:]
```
TheFRedFox: This is totally unnecessary.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
         data['max'] = max(pings)
         data['min'] = min(pings)
 
         send(prefix, data, count)
 
 
 def send(prefix, data, count):
-    template = prefix + '.{}.{} {} ' + str(int(time()))
+    template = str(prefix) + '.{}.{} {} ' + str(int(time.time()))
```
TheFRedFox: Is it ok to have different timestamps here through the whole script run?

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+        data['avg'] = avg
+
+        pings = pings[0:]
         data['max'] = max(pings)
         data['min'] = min(pings)
 
         send(prefix, data, count)
 
 
 def send(prefix, data, count):
```
TheFRedFox: With the change above about the new data structure, `count` parameter is not needed anymore.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
-    template = prefix + '.{}.{} {} ' + str(int(time()))
+    template = str(prefix) + '.{}.{} {} ' + str(int(time.time()))
 
     fails = 0
```
TheFRedFox: What's the reason you calculate fails in here and not above during the `ping` function?

igcollect/ping.py
```diff
@@ -0,0 +1,102 @@
...
+
+        pings = pings[0:]
+        data['max'] = max(pings)
+        data['min'] = min(pings)
+
+        send(prefix, data, count)
+
+
+def send(prefix, data, count):
+    template = str(prefix) + '.{}.{} {} ' + str(int(time.time()))
```
TheFRedFox: `data['dest']` is always the same, you could already put it into the template.

igcollect/ping.py
```diff
@@ -0,0 +1,102 @@
...
+
+    fails = 0
+    for num in range(count):
+        print(template.format(data['dest'], 'ping' + str(num + 1),
+                              data['ping' + str(num)]))
+        if data['ping' + str(num)] == -1:
+            fails += 1
```
TheFRedFox: With the changes about data structure, fails calculation and template we could change it to:
```suggestion
    for idx, ping in enumerate(data['pings'], 1):
        print(template.format('ping' + str(idx), ping))

    fails = dest['fails']
```

igcollect/ping.py
```diff
@@ -0,0 +1,102 @@
...
+        parts = line.split(':')
+        data['dest'] = parts[0].replace(' ', '').replace('.', '_')
+        pings = parts[1].split(' ')
+        # first value would be empty
+        pings.pop(0)
+
+        print(pings)
+        avg = 0
+        for i, ping in enumerate(pings):
+            if pings[i] == '-':
```
TheFRedFox: You are already holding `pings[i]` in `ping` here.

igcollect/ping.py
```diff
@@ -0,0 +1,102 @@
...
+    output = subprocess.getoutput(cmd).split('\n')
+    for line in output:
+        data = {}
+        parts = line.split(':')
+        data['dest'] = parts[0].replace(' ', '').replace('.', '_')
+        pings = parts[1].split(' ')
+        # first value would be empty
+        pings.pop(0)
+
+        print(pings)
```
TheFRedFox: Forgotten 'debug' `print`?

igcollect/ping.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+
+"""igcollect - Latency collection
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+# this script is used to collect ping data to evaluate availability and latency
+# of our network (this should replace smokeping in the longrun)
```
TheFRedFox: `(this should replace smokeping in the longrun` is interesting for the Pull Request itself but not for the script description.

igcollect/ping.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+
+"""igcollect - Latency collection
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+# this script is used to collect ping data to evaluate availability and latency
```
TheFRedFox:
```suggestion
# This script is used to collect ping data to evaluate availability and latency
```

igcollect/ping.py
```diff
@@ -0,0 +1,102 @@
...
+# of our network (this should replace smokeping in the longrun)
+#
+# Please Not that the package fping has to be installed on your Linux server!
```
TheFRedFox:
```suggestion
# Please note that the package fping has to be installed on your linux server!
```

igcollect/ping.py
TheFRedFox: How about using a typo checker before committing to save time for both of us and reduce noise here.

igcollect/ping.py
Jowe1999: fping is much faster than multi threading ping itselve we allready tested it ^^

igcollect/ping.py
Jowe1999: yep, ups :)

igcollect/ping.py
Jowe1999: typo checker allready run 😅

igcollect/ping.py
Jowe1999: can you please explain what you mean or where the error is ?

igcollect/ping.py
Jowe1999: servers.af-control_admin.system.ping.nvidia_com.ping10 162.41 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping11 166.09 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping12 165.95 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping13 163.94 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping14 164.13 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping15 165.00 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping16 200.77 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping17 164.14 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping18 164.89 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping19 164.24 1583400411
servers.af-control_admin.system.ping.nvidia_com.ping20 162.62 1583400411
servers.af-control_admin.system.ping.nvidia_com.fails 0 1583400411
servers.af-control_admin.system.ping.nvidia_com.fails/1 0.0 1583400411
servers.af-control_admin.system.ping.nvidia_com.min 162.41 1583400411
servers.af-control_admin.system.ping.nvidia_com.max 200.77 1583400411
servers.af-control_admin.system.ping.nvidia_com.avg 166.914 1583400411

Jowe1999: amazon.com     : 105.94 102.03 104.78 106.39 102.64 101.89 103.49 102.30 101.32 100.77 102.95 99.68 101.57 102.56 102.27 101.69 101.83 102.35 112.64 102.49
google.de      : 16.90 20.21 18.16 15.86 17.35 17.51 19.87 21.28 18.99 17.88 22.26 18.72 17.36 17.82 18.45 17.38 17.27 20.58 20.71 24.57
bing.de        : 9.11 8.91 7.56 7.52 10.74 7.37 9.52 12.29 9.79 8.69 11.54 10.62 10.37 8.15 7.84 13.30 11.69 10.32 9.26 12.65
duckducogo.com : 18.45 19.54 18.33 18.27 20.35 18.62 20.38 24.27 21.24 20.27 19.30 22.88 23.32 19.42 19.32 20.39 28.69 25.49 21.20 20.33
microsoft.com  : - - - - - - - - - - - - - - - - - - - -
facebook.com   : 4.32 5.28 4.41 4.99 4.55 4.98 6.98 7.75 3.05 4.87 5.15 5.56 5.02 4.21 4.48 4.23 10.80 8.19 5.23 6.73
amazon.de      : 33.72 31.45 32.87 33.61 30.33 29.56 33.51 30.46 32.32 31.19 33.16 32.91 30.55 34.69 31.80 35.02 33.93 32.23 33.01 32.29
bing.com       : 11.50 8.37 10.34 8.53 10.55 10.00 9.73 8.78 8.83 8.15 9.86 9.55 8.57 9.34 8.56 11.63 11.19 11.22 11.99 12.07
facebook.de    : 5.71 5.76 5.23 2.57 6.33 3.99 5.59 6.26 8.04 5.93 5.76 5.82 6.36 11.90 4.84 7.35 5.93 8.32 8.34 7.14
nvidia.com     : 164.35 164.37 165.78 164.13 164.21 163.73 167.73 165.68 165.67 166.29 166.51 166.59 167.34 167.13 165.61 165.80 164.69 167.79 167.22 167.10

igcollect/ping.py
Jowe1999: it's easyer because i can send it directly

igcollect/ping.py
Jowe1999: the server takes it per minute at the end so this makes no big difference

igcollect/ping.py
```diff
@@ -36,21 +36,26 @@ def parse_args():
...
 
 
 def ping(prefix, hosts, count, timeout):
     hosts_string = " ".join(hosts)
     # -p calculates the time fping waits between single ping probes
-    cmd = ('fping -B1 -q -C {} -p {} -t {} {}'.format(count, (20/count) * 1800,
-                                                      timeout, hosts_string))
+    # fping used 20 probes/min by default und set this value to 1800
+    # if you use 30 probes the -p time should goe down, for 10 up
+    # 0 would not make any sense to enter in to this script
```
TheFRedFox: How about stating that on the script parameter itself and not here, so the user knows about it. And how about preventing it directly by checking the parameter? As a user I don't want to get a `DivisionByZero` Exception, but a `You must pass a positive number for 'count'`.

igcollect/ping.py
```diff
@@ -36,21 +36,26 @@ def parse_args():
...
+        p.starmap(ping, data)
+
 
 
 def ping(prefix, hosts, count, timeout):
     hosts_string = " ".join(hosts)
     # -p calculates the time fping waits between single ping probes
-    cmd = ('fping -B1 -q -C {} -p {} -t {} {}'.format(count, (20/count) * 1800,
-                                                      timeout, hosts_string))
+    # fping used 20 probes/min by default und set this value to 1800
```
TheFRedFox: Still doesn't explain why exactly 1800? What makes it important to wait exactly 1800 milliseconds between each try when passing 20.
And what's the reason we don't provide this as a parameter to just pass in? Maybe I am missing something here, but I don't see the point of having such a fixed calculation here?

igcollect/ping.py
TheFRedFox: It's not a problem anymore with the new code. But the code which this comment was related, should not have worked. The calculations were wrong.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+    cmd = ('fping -B1 -q -C {} -p {} -t {} {}'.format(count, (20/count) * 1800,
+                                                      timeout, hosts_string))
+    output = subprocess.getoutput(cmd).split('\n')
     for line in output:
         data = {}
-        part = line.split(':')
-        data['dest'] = part[0].replace(' ', '').replace('.', '_')
-        pings = part[1].split(' ')
-        i = 1
+        parts = line.split(':')
```
TheFRedFox: How about passing a `maxsplit` too as we just want to have one split:
```suggestion
        parts = line.split(':', 1)
```

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+        data['dest'] = parts[0].replace(' ', '').replace('.', '_')
```
TheFRedFox: Was `replace` intentionally used rather than `strip`? Is possible, just going to be sure here.

igcollect/ping.py
```diff
@@ -4,83 +4,90 @@
...
+        pings = parts[1].split(' ')
```
TheFRedFox: You rather want to split by any whitespace and not just by exactly one whitespace. `split` has it as default, so:
```suggestion
        pings = parts[1].split()
```

igcollect/ping.py
TheFRedFox: Thx, how about to put one example line as comment into the code, so the next maintainer will directly know about it?
Like:
```
# example.com : 105.94 - 104.78
# Showing times for each ping. A ping timeout will be shown as '-'.
```

igcollect/ping.py
TheFRedFox: Milliseconds could also result in a minute overlap. How about you either have a global variable holding the current time once at the beginning of the script; or better to get the time once in `main` or in my suggested `check_pings` function and pass it to the `send` function as parameter.

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+
+    for data in values:
+        send(prefix, data, timestamp)
+    
+
+def pings(hosts, count, timeout):
+    hosts_string = " ".join(hosts)
+    # -p calculates the time fping waits between single ping probes
+    # with 20 probes and 1800 milliseconds in between it takes roughly 1 minute
+    # if you use 30 probes the -p time should goe down, for 10 up
```
TheFRedFox: Now we are talking! With this just the middle statement is needed:
```suggestion
    # with 20 probes and 1800 milliseconds in between it takes roughly 1 minute
```
What about if the user is passing a different timeout for the script?

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+    values = pings(hosts, count, timeout)
+
+    timestamp = str(int(time.time()))
+
+    for data in values:
+        send(prefix, data, timestamp)
+    
+
+def pings(hosts, count, timeout):
+    hosts_string = " ".join(hosts)
```
TheFRedFox: Single Quotes!
```suggestion
    hosts_string = ' '.join(hosts)
```

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+    # -p calculates the time fping waits between single ping probes
+    # with 20 probes and 1800 milliseconds in between it takes roughly 1 minute
+    # if you use 30 probes the -p time should goe down, for 10 up
+    cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
+    cmd = cmd.format(count, (20/count) * 1800, timeout, hosts_string)
```
TheFRedFox: To reduce the amount of magic numbers and calculation complexity a bit, I would suggest:
```suggestion
    ttl_factor = 36000  # resulted from tests with 20 probes for under 1 min
    cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
    cmd = cmd.format(count, ttl_factor / count, timeout, hosts_string)
```
With that, you also don't need to explanation above.

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # example.com : 105.94 - 104.78
+    # Showing times for each ping. A ping timeout will be shown as '-'.
+    output = subprocess.getoutput(cmd).split('\n')
+
+    values = []
+    for line in output:
+        data = {}
+        parts = line.split(':', 1)
+        data['dest'] = parts[0].strip(' ').replace('.', '_')
```
TheFRedFox: `' '` is already default for `strip`:
```suggestion
        data['dest'] = parts[0].strip().replace('.', '_')
```

TheFRedFox: Btw, this will not be the same as `replace(' ', '')` from before if `parts[0]` contains spaces inside the hostnames.
Split would make `   am azon.com   ` to `am azon.com` and not `amazon.com`.
If this is a problem, rather use `replace`, therefore the question before.

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+        data['min'] = min(pings)
+        
+        total = 0
+        fails = 0
+        for i, ping in enumerate(pings):
+            if ping == '-':
+                pings[i] = -1
+                fails += 1
+            else:
+                total += float(pings[i])
```
TheFRedFox: You already have `pings[i]`:
```suggestion
                total += float(ping)
```

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+    for line in output:
+        data = {}
+        parts = line.split(':', 1)
+        data['dest'] = parts[0].strip(' ').replace('.', '_')
+        pings = parts[1].split()
+
+        data['max'] = max(pings)
+        data['min'] = min(pings)
+        
+        total = 0
```
TheFRedFox: As you are awaiting floats, make it already clear here:
```suggestion
        total = 0.0
```

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+        data['fails'] = fails
+
+        values.append(data)
+
+    return values
+
+
+def send(prefix, data, timestamp):
+    template = \
+        str(prefix) + '.' + data['dest'] + '.{} {} ' + timestamp
```
TheFRedFox: PEP8, no use of `\`:
```suggestion
    template = (
        str(prefix) + '.' + data['dest'] + '.{} {} ' + timestamp
    )
```

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+
+    for num in range(len(data['pings'])):
+        print(template.format('ping' + str(num + 1), data['pings'][num]))
```
TheFRedFox:
```suggestion
    for num, ping in enumerate(data['pings'], 1):
        print(template.format('ping' + str(num), ping))
```

igcollect/ping.py
Jowe1999: I make it adjustable

TheFRedFox: If it is needed, ok. Therefore the question, maybe it is not needed.

igcollect/ping.py
Jowe1999: thats what i want i checked it

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+
+    print(template.format('fails', data['fails']))
+    print(template.format('fails/1', data['fails'] / len(data['pings'])))
```
TheFRedFox: Theoretically possible that `len` will be `0`. Results to `DivideByZero` error. Practically I guess it won't happen though. What's the possibility for you?

igcollect/ping.py
```diff
@@ -0,0 +1,122 @@
...
+            data['avg'] = avg
+            
+        data['fails'] = fails
+
+        values.append(data)
+
+    return values
+
+
+def send(prefix, data, timestamp):
```
TheFRedFox: Btw, I guess `print_pings` would be a better naming, as the function not really sends something. Really minor though.

igcollect/ping.py
TheFRedFox: Good. Then go for it.

igcollect/ping.py
Jowe1999: when? even if no ping is responding it would be 20

TheFRedFox: Theoretically, if `data` was passed like that. Practically it should not happen. One could make a check though just for defensive programming purpose. Not needed though.

Jowe1999: ok you'r right

igcollect/ping.py
```diff
@@ -0,0 +1,134 @@
...
+                            socket.gethostname().replace('.', '_')),
+                        help='the path to the value in Graphite '
+                             '(default= servers.(hostname).system.ping)')
+    parser.add_argument('-t', '--timeout', type=int, default=300,
+                        help='time in milliseconds '
+                             'till the timeout is retched (default= 300)')
+    parser.add_argument('-d', '--delay', type=int, default=1800,
+                        choices=range(0,5000),
+                        help='time to wait between a series of pings '
+                             '(default= 1800)')
```
TuxPowered42: 1800 of what?

igcollect/ping.py
```diff
@@ -0,0 +1,134 @@
...
+                             '(number has to be bigger '
+                             'than 1 and smaller than 60) (default= 20)')
+    parser.add_argument('-p', '--prefix',
+                        default='servers.{}.system.ping'.format(
+                            socket.gethostname().replace('.', '_')),
+                        help='the path to the value in Graphite '
+                             '(default= servers.(hostname).system.ping)')
+    parser.add_argument('-t', '--timeout', type=int, default=300,
+                        help='time in milliseconds '
+                             'till the timeout is retched (default= 300)')
```
TuxPowered42: https://en.wiktionary.org/wiki/retch#English

Description of timeout here differs from the one in code below, where it's described as delay and it does not explain at all what is it for.

igcollect/ping.py
```diff
@@ -0,0 +1,134 @@
...
+        str(prefix) + '.' + data['dest'] + '.{} {} ' + timestamp
+    )
+
+    for num, ping in enumerate(data['pings'], 1):
+        print(template.format('ping' + str(num), ping))
+
+    print(template.format('fails', data['fails']))
+
+    if len(data['pings']) == 0:
+        raise Exception('something went horrible wrong - check your fping -')
```
TuxPowered42: What's with those hyphens?

igcollect/ping.py
```diff
@@ -0,0 +1,134 @@
...
+
+    values = []
+    for line in output:
+        data = {}
+        parts = line.split(':', 1)
+        data['dest'] = parts[0].strip(' ').replace('.', '_')
+        pings = parts[1].split()
+
+        data['max'] = max(pings)
+        data['min'] = min(pings)
```
TuxPowered42: I still want standard deviation calculated.

igcollect/ping.py
Jowe1999: ms I'll add this

igcollect/ping.py
Jowe1999: ok I'll add

igcollect/ping.py
Jowe1999: these are different values !

igcollect/ping.py
```diff
@@ -0,0 +1,141 @@
...
+                        help='the path to the value in Graphite '
+                             '(default= servers.(hostname).system.ping)')
+    parser.add_argument('-t', '--timeout', type=int, default=500,
+                        help='Initial target timeout in milliseconds. '
+                             'the amount of time that program '
+                             'waits for a response (default= 500)')
+    parser.add_argument('-d', '--delay', type=int, default=1800,
+                        choices=range(0,5000),
+                        help='time to wait between a series of pings in '
+                             'milliseconds (default= 1800 for 20 pings)')
```
TuxPowered42: You're duplicating defaults everywhere. Use https://docs.python.org/3/library/argparse.html#argparse.ArgumentDefaultsHelpFormatter instead.

igcollect/ping.py
```diff
@@ -0,0 +1,141 @@
...
+    parser.add_argument('hosts', nargs='+',
+                        help='the hosts to ping')
+    parser.add_argument('-c', '--count', type=int, default=20,
+                        choices=range(1,60),
+                        help='how many times the script will try to ping '
+                             '(number has to be bigger '
+                             'than 1 and smaller than 60) (default= 20)')
+    parser.add_argument('-p', '--prefix',
+                        default='servers.{}.system.ping'.format(
+                            socket.gethostname().replace('.', '_')),
```
TuxPowered42: Please see other collectors. We only provide simple prefix, the correct one is to be configured by puppet.

igcollect/ping.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python3
+
+"""igcollect - Latency collection
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+# This script is used to collect ping data to evaluate availability and latency
+# of our network and make it easy to create a Graphite Dashboard
+#
+# Please note that the package fping has to be installed on your Linux server!
```
kofrezo: - [ ] This belongs to the doc string above
- [ ] Do not forget the license

igcollect/ping.py
```diff
@@ -0,0 +1,137 @@
...
+                     for i in range(0, len(args.hosts), 10)]
+
+    data = [(args.prefix, chunk, args.count, args.timeout, args.delay)
+            for chunk in chunked_hosts]
+
+    with Pool(cpu_count() * 2) as p:
+        p.starmap(check_pings, data)
+
+
+def check_pings(prefix, hosts, count, timeout, delay):
```
kofrezo: - [ ] Do not forget the doc strings for the functions

igcollect/ping.py
```diff
@@ -0,0 +1,137 @@
...
+
+    timestamp = str(int(time.time()))
+
+    for data in values:
+        print_ping(prefix, data, timestamp)
+
+
+def pings(hosts, count, timeout, delay):
+    hosts_string = ' '.join(hosts)
+    cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
```
kofrezo: - [ ] Keep in mind that if you run the script in a non shell (no or limited path variable) you have to write the full path e.g. /usr/bin/fping

igcollect/ping.py
TheFRedFox: The original reply is still valid.

igcollect/ping.py
```diff
@@ -0,0 +1,137 @@
...
+        total = 0.0
+        fails = 0
+        for i, ping in enumerate(pings):
+            if ping == '-':
+                pings[i] = '-1'
+                fails += 1
+            else:
+                total += float(ping)
+
+        pings = list(map(float, pings))
```
TheFRedFox:
```suggestion
                pings[i] = -1.0
                fails += 1
            else:
                ping[i] = float(ping)
                total += float(ping)
```

igcollect/ping.py
```diff
@@ -0,0 +1,137 @@
...
+
+    return values
+
+
+def print_ping(prefix, data, timestamp):
+    template = (
+        str(prefix) + '.' + data['dest'] + '.{} {} ' + timestamp
+    )
+
+    for num, ping in enumerate(data['pings'], 1):
```
TheFRedFox: You are using `data['pings']` quite often over next lines. How about we store that first into another variable?

igcollect/ping.py
```diff
@@ -0,0 +1,137 @@
...
+        print(template.format('ping' + str(num), ping))
+
+    print(template.format('fails', data['fails']))
+
+    if len(data['pings']) == 0:
```
TheFRedFox:
```suggestion
    if not data['pings']:
```

igcollect/ping.py
```diff
@@ -0,0 +1,137 @@
...
+        raise Exception('something went horrible wrong: check your fping')
```
TheFRedFox: **First**:
```suggestion
        raise Exception('Something went horribly wrong: Check your fping!')
```

**Second**: Do not use so generic Exceptions. Do not make the user of the function in need of catching such a broad Exception. There are quite some good Exceptions already in Python which you can use for this occasion here. Make use of them.

**Third**: We made this function to be more of a utility function. Make the messages as part of such a utility function. The utility function does not know about fping or whatever was used. For this function the parameter which were given doesn't make sense.

**Fourth**: Actually for me it does makes sense. One could also just print a part of the data and just skip the other one. So for me if would make sense to just skip the print if the length is `0`.

igcollect/ping.py
Jowe1999: would make it much simpler to read 👍

igcollect/ping.py
```diff
@@ -113,24 +144,35 @@ def pings(hosts, count, timeout, delay):
...
-    for num, ping in enumerate(data['pings'], 1):
+    for num, ping in enumerate(pings, 1):
         print(template.format('ping' + str(num), ping))
 
     print(template.format('fails', data['fails']))
 
-    if len(data['pings']) == 0:
-        raise Exception('something went horrible wrong: check your fping')
+    if len(pings):
+        # If the length is 0 we ignore the entries
```
TheFRedFox: What is the reason to ignore all entries? The other entries are possible to print.

igcollect/ping.py
```diff
@@ -113,24 +144,35 @@ def pings(hosts, count, timeout, delay):
...
 
-    for num, ping in enumerate(data['pings'], 1):
+    for num, ping in enumerate(pings, 1):
         print(template.format('ping' + str(num), ping))
 
     print(template.format('fails', data['fails']))
 
-    if len(data['pings']) == 0:
-        raise Exception('something went horrible wrong: check your fping')
+    if len(pings):
```
TheFRedFox:
```suggestion
    if pings:
```

igcollect/ping.py
```diff
@@ -113,24 +144,35 @@ def pings(hosts, count, timeout, delay):
 
 
 def print_ping(prefix, data, timestamp):
+    """ prints the data in a format we can collect
+
+        Parameters:
+            prefix (str): were the script will put the data in graphite
+            data (dict(floar)): the data gathered (pings, min, may, avg, std)
```
TheFRedFox:
```suggestion
            data (dict(float)): the data gathered (pings, min, max, avg, std)
```

igcollect/ping.py
TheFRedFox: As for utility functions it is absolutely meaningful to have, I can't really see the benefit for the `main` and `parse_args` function. The `main` function should be explain in the script description already and the `parse_args` function partly already in the script explanation as for how to use the script and also mostly redundant if the `parse_args` is not complex.
What is your opinion about that @kofrezo?

igcollect/ping.py
```diff
@@ -58,6 +67,17 @@ def main():
 
 
 def check_pings(prefix, hosts, count, timeout, delay):
+    """ checks the pings and prints them
+
+        Parameters:
+            prefix (str): were the script will put the data in graphite
```
TheFRedFox:
```suggestion
            prefix (str): where the script will put the data in graphite
```
however:
```suggestion
            prefix (str): the prefix for the output for graphite
```

igcollect/ping.py
Jowe1999: we just ignore the entries of the not send stuff every other thing will be written

igcollect/ping.py
kofrezo: Agreed. I would not complain if there is no doc string for the main and parse_args function but everything else should have one.

igcollect/ping.py
```diff
@@ -0,0 +1,179 @@
...
+            values (list(dict(float))): all data gathered for all pinged hosts
+    """
+
+    hosts_string = ' '.join(hosts)
+    cmd = 'fping -B1 -q -C {} -p {} -t {} {}'
+    cmd = cmd.format(count, delay, timeout, hosts_string)
+    # output:
+    # example.com : 105.94 - 104.78
+    # Showing times for each ping. A ping timeout will be shown as '-'.
+    output = subprocess.getoutput(cmd).split('\n')
```
TuxPowered42, line 108: Please switch to `subprocess.check_output()`, so that a shell is not spawned. Running things in shell is a bad security practice.

PR #134, opened 2020-03-10
Add a script to collect statuscodes from nginx and similar services

Jowe1999:
This script (statuscodes2graphite) scans the log file than fetches and counts all stuscodes and sends the data to graphite.

### DIFF
--- a/igcollect/statuscode2gaphite.py
+++ b/igcollect/statuscode2gaphite.py
@@ -1,0 +1,108 @@
+import argparse
+import datetime
+import time
+import socket
+
+from collections import Counter
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        description='Script to collect statuscodes from nginx')
+    parser.add_argument('logfile',
+                        help='path to logfile')
+    parser.add_argument('-s', '--software', default='nginx',
+                        help='which tool is using it ')
+    parser.add_argument('-t', '--time', type=int, default=3,
+                        help='where to search for the time in the log file ')
+    parser.add_argument('-x', '--position', type=int, default=8,
+                        help='where to search for the code in the log file ')
+    parser.add_argument('-p', '--prefix',
+                        default='servers.{}.software.{}'.format(
+                            socket.gethostname().replace('.', '_'),
+                            parser.parse_args().software),
+                        help='the path to the value in Graphite ')
+    return parser.parse_args()
+
+
+def get_position(position, entry):
+    return entry[position]
+
+
+def get_lines(path):
+    log_file = open(path)
+    log = []
+    for line in log_file:
+        log.append(line.split(' '))
+    return log
+
+
+def count_uniq_statuscodes(log, position):
+    statuscodes = []
+    for entry in log:
+        statuscodes.append(entry[position])
+    statuscodes_sorted = dict(Counter(statuscodes))
+
+    return statuscodes_sorted, len(statuscodes)
+
+
+def add_seconds_to_time(time, seconds):
+    new_time = time + datetime.timedelta(seconds=seconds)
+    return new_time.strftime('%d/%b/%Y:%H:%M:%S')
+
+
+def find_row(log, timedate_at_start, seconds_to_add, time_position):
+    for number_of_entry, entry in enumerate(log):
+        search_time = add_seconds_to_time(timedate_at_start, seconds_to_add)
+        if search_time in get_position(time_position, entry):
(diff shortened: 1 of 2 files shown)


igcollect/statuscode2gaphite.py
```diff
@@ -0,0 +1,108 @@
+import argparse
```
TuxPowered42: Missing shebang and copyright.

igcollect/statuscode2gaphite.py
```diff
@@ -0,0 +1,108 @@
...
+                        help='path to logfile')
+    parser.add_argument('-s', '--software', default='nginx',
+                        help='which tool is using it ')
+    parser.add_argument('-t', '--time', type=int, default=3,
+                        help='where to search for the time in the log file ')
+    parser.add_argument('-x', '--position', type=int, default=8,
+                        help='where to search for the code in the log file ')
+    parser.add_argument('-p', '--prefix',
+                        default='servers.{}.software.{}'.format(
+                            socket.gethostname().replace('.', '_'),
```
TuxPowered42: Please see other collectors. We only provide simple prefix, the correct one is to be configured by puppet.

igcollect/statuscode2gaphite.py
```diff
@@ -0,0 +1,108 @@
...
+def get_position(position, entry):
+    return entry[position]
+
+
+def get_lines(path):
+    log_file = open(path)
+    log = []
+    for line in log_file:
+        log.append(line.split(' '))
+    return log
```
TuxPowered42: This loads the whole log to memory. Can't it be done using streams?

Jowe1999: this is already streaming :D

igcollect/statuscode2gaphite.py
Jowe1999: Is this really what you want if someone forgets to configure it right to just throw it to nginx.(etc)

TuxPowered42: We don't forget to configure the prefix because we have Puppet helper for it.

igcollect/statuscode2gaphite.py
TuxPowered42: No, it's not. You're allocating an array and pushong the whole file to it, line by line. If the file is a few hundred MiB, this script will consume this much memory.

Jowe1999:
I testet the script on de7w1.foe and it worked quick and reliable

igcollect/statuscode2gaphite.py
Jowe1999: https://docs.python.org/3.5/library/io.html

Jowe1999: The problem at the end is to convert the text to a set of lists ^^

TuxPowered42: You could do it in a streaming way, count codes as you go through the log file line by line.

Jowe1999: even on a forge server thi script uses a max of 1.1mb ram / run

TuxPowered42: I would prefer to not take any chances here. This script is potentially disasterous.

igcollect/statuscode2graphite.py
```diff
@@ -97,23 +89,31 @@ def get_log_for_period(log, timedate_at_start, period_in_sec, time_position):
...
 
-    for code in codes.keys():
-        print(template.format('.status_codes', code, codes[code], prob_time))
+def add_seconds_to_time(time, seconds):
+    new_time = time + datetime.timedelta(seconds=seconds)
+    return new_time.strftime('%d/%b/%Y:%H:%M:%S')
 
-    print(template.format('', 'requests', total, prob_time))
+
+def find_row(log, timedate_at_start, seconds_to_add, time_position):
```
TuxPowered42, line 110: After dicussion on Slack I ask for a bit of documentation why this non-obious way is faster than single scan over the file.

igcollect/statuscode2graphite.py
```diff
@@ -97,23 +89,31 @@ def get_log_for_period(log, timedate_at_start, period_in_sec, time_position):
...
 
-def main():
-    args = parse_args()
-    prob_time = int(time.time())
-    search_time = datetime.datetime.utcnow()
+def count_uniq_statuscodes(log, position):
+    statuscodes = []
+    for entry in log:
+        statuscodes.append(entry[position])
+    statuscodes_sorted = dict(Counter(statuscodes))
```
TuxPowered42, line 96: There is nothing sorted about `dict`. Why the conversion at all?

PR #135, opened 2020-03-10
Gather stats from Strongswan VPN server

TuxPowered42:


igcollect/strongswan.py
```diff
@@ -0,0 +1,57 @@
...
+    # ---- >8 ----
+
+    # Let's say that the line containing word 'remote' and IP address of the client
+    # identifies a single client.
+    line_re = re.compile(r'^\s+remote .* @')
+
+
+    for line in check_output(['/usr/sbin/swanctl', '--list-sas']).splitlines():
+        if line_re.match(line.decode()):
+            clients += 1
```
lamaral: I think you can replace this for loop with a `line_re.finditer()`, if you add `re.MULTILINE` to the compile arguments.

igcollect/strongswan.py
```diff
@@ -0,0 +1,57 @@
...
+    #     local  0.0.0.0/0 ::/0
+    #     remote XXX.XXX.XXX.XXX/32 XXX:XXX:XXX::XXX/128
+    # ---- >8 ----
+
+    # Let's say that the line containing word 'remote' and IP address of the client
+    # identifies a single client.
+    line_re = re.compile(r'^\s+remote .* @')
+
+
+    for line in check_output(['/usr/sbin/swanctl', '--list-sas']).splitlines():
```
TheFRedFox: I would suggest to write it as that:
``` python
output = check_output(...).decode()
lines = output.splitlines()
clients = sum(1 for l in lines if line_re.match(l))
```

igcollect/strongswan.py
```diff
@@ -0,0 +1,57 @@
...
+    #     installed 3354s ago, rekeying in 53s, expires in 606s
+    #     in  42424242 (-|0x000675ec), 9577036 bytes, 63405 packets,     2s ago
+    #     out 42424242 (-|0x000675ec), 20466058 bytes, 68523 packets,     2s ago
+    #     local  0.0.0.0/0 ::/0
+    #     remote XXX.XXX.XXX.XXX/32 XXX:XXX:XXX::XXX/128
+    # ---- >8 ----
+
+    # Let's say that the line containing word 'remote' and IP address of the client
+    # identifies a single client.
+    line_re = re.compile(r'^\s+remote .* @')
```
TheFRedFox: I do it correctly, that you just want to match the 4th line, right? Not the 12th line.

igcollect/strongswan.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def count_connected_clients():
+    clients = 0
+    
+    # Single connected users looks like this:
+
+    # ---- 8< ----
+    # vpn_by_id_group_xyz: #974, ESTABLISHED, IKEv2, 4242424242424242_i 4242424242424242_r*
+    #   local  'XXXXXX.innogames.de' @ XXX.XXX.XXX.XXX[4500]
```
TheFRedFox: Maybe rather go for `'XXXXX.example.com'` for the example?

igcollect/strongswan.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
+"""igcollect - Strongswan stats
+
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+from argparse import ArgumentParser
+from subprocess import check_output
+from time import time
```
TheFRedFox: You are shadowing the module `time` with that import. For me it is fine, but in the future one needs to think about it.

igcollect/strongswan.py
```diff
@@ -0,0 +1,57 @@
...
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='strongswan')
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    template = args.prefix + '.{} {} ' + str(int(time()))
+    print(template.format('clients', count_connected_clients()))
```
TheFRedFox: If `clients` is fixed, we could also already put it in the template.

Woellchen:
@TuxPowered42 any progress?

igcollect/strongswan.py
TuxPowered42: Fixed.

igcollect/strongswan.py
TuxPowered42: I expect more stats to be added in future.

igcollect/strongswan.py
TuxPowered42: Fixed.

igcollect/strongswan.py
TuxPowered42: Line 12 has no `@` sign in it, which is matched by this regexp.

igcollect/strongswan.py
TuxPowered42: I rewrote it the way @TheFRedFox suggested, sorry.

igcollect/strongswan.py
TuxPowered42: As discussed on Slack, I consider shadowing good for this simple script.

PR #136, opened 2020-04-22
Fix logvalues with empty TZ environment

Felixoid:
`utcnow` returns the wrong values for some reason. From the documentation https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow :

```
This is like now(), but returns the current UTC date and time, as a naive datetime object. An aware current UTC datetime can be obtained by calling datetime.now(timezone.utc). See also now().
```

### DIFF
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -23,7 +23,7 @@ import time
 import os
 import gzip
 import logging
-from datetime import datetime, timezone
+import datetime
 
 from os.path import exists
 from argparse import ArgumentParser, ArgumentTypeError
@@ -50,7 +50,8 @@ class Metric:
         self.period = period
         self.values = []  # Container for metric values
         self.last_value = 0
-        self.now = int(datetime.now(timezone.utc).timestamp())
+        self.now = int(datetime.datetime.now(datetime.timezone.utc)
+                       .timestamp())
 
     def get_timeshift(self):
         if self.period:
@@ -258,7 +259,7 @@ def convert_to_timestamp(time_str, time_format):
         if time_format.endswith('z') and time_str[-3] == ':':
             time_str = ''.join(time_str.rsplit(':', 1))
 
-        dt = datetime.strptime(time_str, time_format).utctimetuple()
+        dt = datetime.datetime.strptime(time_str, time_format).utctimetuple()
         timestamp = time.mktime(dt)
     except ValueError:
         timestamp = int(time_str)
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -23,7 +23,7 @@ import time
 import os
 import gzip
 import logging
-import datetime
+from datetime import datetime, timezone
 
 from os.path import exists
 from argparse import ArgumentParser, ArgumentTypeError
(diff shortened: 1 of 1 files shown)


igcollect/logfile_values.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, arg):
         self.period = period
         self.values = []  # Container for metric values
         self.last_value = 0
-        self.now = int(datetime.datetime.utcnow().timestamp())
+        self.now = int(datetime.now(timezone.utc).timestamp())
```
TheFRedFox, line 53: Is the specification for utc necessary anyway? Shouldn't `timestamp` already give you the "utc" timestamp? Different timezones doesn't change the timestamp, no?

igcollect/logfile_values.py
```diff
@@ -23,7 +23,7 @@
 import os
 import gzip
 import logging
-import datetime
+from datetime import datetime, timezone
```
TheFRedFox, line 26: Arguable, as you like that shadow the `datetime` root module. With the `datetime` import.

igcollect/logfile_values.py
Felixoid: I'm not sure, just following the documentation and trying to leave the desired behavior. Unfortunately, right now it's just broken =\

Felixoid: `flake8 igcollect/logfile_values.py` keeps silence

TheFRedFox: Not sure what `flake8` has to do with that. Made a quick test:
```
>>> print(int(datetime.datetime.now(datetime.timezone.utc).timestamp()))
1587572110
>>> print(int(datetime.datetime.now().timestamp()))
1587572111
```
It's the same. `utcnow` is different though, that's correct. So the fix is correct, but `datetime.timezone.utc` is not necessary as parameter.

igcollect/logfile_values.py
TheFRedFox: `Line too long` I would not accept as reason. Stating, that we will not likely add anything else and the script no disadvantage about shadowing it, might be one.
Making just aware of it as it might be dangerous and can be still ok to do.

igcollect/logfile_values.py
```diff
@@ -50,7 +50,8 @@ def __init__(self, arg):
         self.period = period
         self.values = []  # Container for metric values
         self.last_value = 0
-        self.now = int(datetime.datetime.utcnow().timestamp())
+        self.now = int(datetime.datetime.now(datetime.timezone.utc)
+                       .timestamp())
```
TheFRedFox, line 54:
```suggestion
        self.now = int(datetime.datetime.now().timestamp())
```
Should be the same.

TheFRedFox:
Just mentioned the redundant `datetime.timezone.utc`. No rejection though.

igcollect/logfile_values.py
Felixoid: I partially agree, but that's the statement of the utcnow() documentation

```
An aware current UTC datetime can be obtained by calling datetime.now(timezone.utc)
```

And as I like to simplify everything, I just don't like kidding with time =)

PR #137, opened 2020-04-23
Another place with time w/o aware about TZ

Felixoid:
`utctimetuple` is another `don't use it` place https://docs.python.org/3/library/datetime.html#datetime.datetime.utctimetuple

Follows #136

### DIFF
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -161,7 +161,11 @@ def parse_args():
     parser.add_argument('--columns-num', default='0', type=int)
     parser.add_argument('--metric', type=Metric, nargs='+')
     parser.add_argument('--time-column', default='0', type=int)
-    parser.add_argument('--time-format', default='%Y-%m-%dT%H:%M:%S%z')
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


igcollect/logfile_values.py
```diff
@@ -259,7 +259,7 @@ def convert_to_timestamp(time_str, time_format):
         if time_format.endswith('z') and time_str[-3] == ':':
             time_str = ''.join(time_str.rsplit(':', 1))
 
-        dt = datetime.datetime.strptime(time_str, time_format).utctimetuple()
+        dt = datetime.datetime.strptime(time_str, time_format).timetuple()
```
TheFRedFox, line 262: I tested around a bit. And I still see problems with `timetuple` if the timezone is not specified, which could happen if `time_format` does not end with `z`.
We need to raise the question, what should happen if `strptime` does not return a datetime with a timezone.
Furthermore, I also see less problems with just using the `timestamp` method of the `datetime` object.
A solution could be to check if the datetime has a timezone, if not, we set the timezone explicitly to `utc` and then just using the `timestamp` method of the datetime object.

PR #140, opened 2020-05-07
Set R6515 to the same total cpu perf as M640

raimund-s:
Set AMD's 7502P 32 cores to be equivalent to the 40 cores of a dual socket Xeon 6148 system.


TheFRedFox:
Don't forget the Copyright for 2020.

raimund-s:
> Don't forget the Copyright for 2020.

year updated

PR #145, opened 2020-07-30
Gather stats from /proc/pressure

meikomeik:
Gather stats from /proc/pressure/*

https://www.kernel.org/doc/html/latest/accounting/psi.html

### DIFF
--- a/igcollect/linux_pressure.py
+++ b/igcollect/linux_pressure.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+"""igcollect - PSI - Pressure Stall Information
+
+Copyright (c) 2020 InnoGames GmbH
+"""
+
+import os
+import time
+
+from argparse import ArgumentParser
+
+
+def parse_args():
+    parser = ArgumentParser(prog='linux_pressure.py')
+    parser.add_argument('--prefix', default='pressure')
+    return parser.parse_args()
+
+
+def main():
+    if not os.path.isdir('/proc/pressure'):
+        return
+
+    args = parse_args()
+    for pressure in get_pressure_stats():
+        print('{}.{}'.format(args.prefix, pressure))
+
+
+def parse_split_file(filename):
+    """Utility to read and parse a space delimited file"""
+    with open(filename, 'r') as fd:
+        return [line.split() for line in fd]
+
+
+def get_pressure_stats():
+    """/proc/pressure/{cpu,memory,io} aggregates the pressure information for
+    each of the resource: cpu, memory, and io.
+
+    The format for CPU is as such:
+        some avg10=0.00 avg60=0.00 avg300=0.00 total=0
+    and for memory and IO:
+        some avg10=0.00 avg60=0.00 avg300=0.00 total=0
+        full avg10=0.00 avg60=0.00 avg300=0.00 total=0
+
+    The "some" line indicates the share of time in which at least some tasks
+    are stalled on a given resource.
+
+    The "full" line indicates the share of time in which all non-idle tasks are
+    stalled on a given resource simultaneously. In this state actual CPU cycles
+    are going to waste, and a workload that spends extended time in this state
+    is considered to be thrashing. This has severe impact on performance, and
+    it’s useful to distinguish this situation from a state where some tasks are
+    stalled but the CPU is still doing productive work.
+    As such, time spent in this subset of the stall state is tracked separately
+    and exported in the “full” averages.
+    """
+    now = int(time.time())
+    output = []
+    for key in ['cpu', 'io', 'memory']:
+        lines = parse_split_file('/proc/pressure/{}'.format(key))
+        for line in lines:
+            for element in line[1:]:
+                name, value = element.split('=', 1)
+                output.append('{}.{}.{} {} {}'.format(
+                    key, line[0], name, value, now))
+    return output
+
+
+if __name__ == '__main__':
+    main()


igcollect/linux_pressure.py
```diff
@@ -0,0 +1,46 @@
...
+    if os.path.isdir('/proc/pressure'):
+        args = parse_args()
+        for pressure in get_pressure_stats():
+            print('{}.{}'.format(args.prefix, pressure))
+
+
+def parse_split_file(filename):
+    """Utility to read and parse a space delimited file"""
+    with open(filename, 'r') as fd:
+        return [line.strip().split(' ') for line in fd]
```
TheFRedFox, line 29: How long can that file be? If too long, you might want to consider returning a generator instead of a list by replacing the brackets to braces:
```suggestion
        return (line.strip().split(' ') for line in fd)
```

TheFRedFox, line 29: The default of `split()` is already splitting on all whitespaces. Your `split(' ')` instead would just split on one single whitespace. So `'f  f'` would become `['f', '', 'f']`. By using the default `split()` you also don't need the `strip` beforehand. So you could change it to:
```suggestion
        return [line.split() for line in fd]
```

igcollect/linux_pressure.py
```diff
@@ -0,0 +1,46 @@
...
+    parser = ArgumentParser(prog='linux_pressure.py')
+    parser.add_argument('--prefix', default='pressure')
+    return parser.parse_args()
+
+
+def main():
+    if os.path.isdir('/proc/pressure'):
+        args = parse_args()
+        for pressure in get_pressure_stats():
+            print('{}.{}'.format(args.prefix, pressure))
```
TheFRedFox, line 23: Don't be afraid of early returns:

```suggestion
    if not os.path.isdir('/proc/pressure'):
        return
    
    args = parse_args()
    for pressure in get_pressure_stats():
        print('{}.{}'.format(args.prefix, pressure))
```

igcollect/linux_pressure.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def parse_split_file(filename):
+    """Utility to read and parse a space delimited file"""
+    with open(filename, 'r') as fd:
+        return [line.strip().split(' ') for line in fd]
+
+
+def get_pressure_stats():
```
TheFRedFox, line 32: Give the reviewer a comment about how that file or the lines looks like, so the reviewer is actually able to review the parsing process.

igcollect/linux_pressure.py
meikomeik: Size in bytes:
> 56 /proc/pressure/cpu
> 110 /proc/pressure/io
> 110 /proc/pressure/memory
> 276 total

I consider this rather short.

meikomeik: Yes, you are right. Thank you, I used your shorter version now (and tested it. works!)

igcollect/linux_pressure.py
meikomeik: Your not nested version is nicer! I agree. I adopted that code

igcollect/linux_pressure.py
meikomeik: I left a rather lengthy comment (stolen from the documentation) on what /proc/pressure/* is and how the data format looks like.

igcollect/linux_pressure.py
meikomeik: I tried your suggestion:
> ` return (line.split() for line in fd)`
and it breaks on:
>    File "linux_pressure.py", line 29, in <genexpr>
>     return (line.split() for line in fd)
> ValueError: I/O operation on closed file

will keep the code inside square brakets

TheFRedFox: Yeah, 276 bytes is not interesting at all. You can ignore that then. It's more interesting if it's about hundreds of MB, then to not overload the memory with it.
The error makes sense, the file descriptor `fd` is in the generator, if we would need the generator, then we would need to change the process there.
But like I said, 276 bytes are totally ignorable.

igcollect/linux_pressure.py
TheFRedFox: General idea is: Indent the exceptional/not interesting/not important code, and leave the normal/interesting/important code as near to the root indention as possible.

igcollect/linux_pressure.py
```diff
@@ -0,0 +1,69 @@
...
+    """/proc/pressure/{cpu,memory,io} aggregates the pressure information for
+    each of the resource: cpu, memory, and io.
+
+    The format for CPU is as such:
+        some avg10=0.00 avg60=0.00 avg300=0.00 total=0
+    and for memory and IO:
+        some avg10=0.00 avg60=0.00 avg300=0.00 total=0
+        full avg10=0.00 avg60=0.00 avg300=0.00 total=0
+
+    The “some” line indicates the share of time in which at least some tasks
```
TheFRedFox, line 44:
```suggestion
    The "some" line indicates the share of time in which at least some tasks
```

igcollect/linux_pressure.py
```diff
@@ -0,0 +1,69 @@
...
+    are stalled on a given resource.
+
+    The “full” line indicates the share of time in which all non-idle tasks are
```
TheFRedFox, line 47:
```suggestion
    The "full" line indicates the share of time in which all non-idle tasks are
```

igcollect/linux_pressure.py
```diff
@@ -0,0 +1,69 @@
...
+    stalled on a given resource simultaneously. In this state actual CPU cycles
+    are going to waste, and a workload that spends extended time in this state
+    is considered to be thrashing. This has severe impact on performance, and
+    it’s useful to distinguish this situation from a state where some tasks are
+    stalled but the CPU is still doing productive work.
+    As such, time spent in this subset of the stall state is tracked separately
+    and exported in the “full” averages.
```
TheFRedFox, line 54:
```suggestion
    and exported in the "full" averages.
```

igcollect/linux_pressure.py
TheFRedFox: Thx for the descriptive comment. The first part about the format is the most interesting for the reviewer. The second part looks more interesting for the whole script itself, for the single function a short description of how the data will be returned would be more interesting. JFTK.

I leave it resolved, not important for change.

igcollect/linux_pressure.py
meikomeik: copy&paste fail. fixed it

igcollect/linux_pressure.py
meikomeik: copy&paste fail. fixed it

igcollect/linux_pressure.py
meikomeik: copy&paste fail. fixed it

igcollect/linux_pressure.py
Felixoid: Sorry, here it's still back/forth double quotes

PR #148, opened 2020-10-13
Fix py3 compatibility for zookeeper script

Felixoid:


igcollect/zookeeper.py
```diff
@@ -39,19 +39,19 @@ def main():
...
     conn = socket(AF_INET, SOCK_STREAM)
     conn.connect((hostname, port))
-    conn.sendall(content)
+    conn.sendall(content.encode())
     conn.shutdown(SHUT_WR)
     while True:
-        data = conn.recv(1024)
-        if data == '':
+        answer = conn.recv(1024)
+        if answer == b'':
```
TheFRedFox, line 50: Do we need the explicit comparison or could we also go with:
```suggestion
        if not answer:
```
?

TheFRedFox (approved):
For string/byte compatibility this looks good.

kofrezo (approved):

PR #150, opened 2020-12-16
Extend regex for nvme disks

konstantinbest:
Our current regex is not matching with the disk names in AWS (nvme) and therefore
we don't have any disk metrics in graphite. This change will fix this issue.

### DIFF
--- a/igcollect/linux_disk.py
+++ b/igcollect/linux_disk.py
@@ -103,7 +103,7 @@ def get_diskstats_dict():
             x = line.strip().split()
             disk_name = x[2]
             # Filter for only normal disks
-            if not match('^(xv|[shv])d[a-z]?$', disk_name):
+            if not match('^(xv|[shv]|nv)(d[a-z]|me[0-9]n[1-9])$', disk_name):
                 continue
             ret[disk_name] = {'type': disk_type}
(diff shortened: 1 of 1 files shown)


igcollect/linux_disk.py
```diff
@@ -103,7 +103,7 @@ def get_diskstats_dict():
             x = line.strip().split()
             disk_name = x[2]
             # Filter for only normal disks
-            if not match('^(xv|[shv])d[a-z]?$', disk_name):
+            if not match('^(xv|[shv]|nv)[dm][a-z]?[0-9]?n?[1-9]?$', disk_name):
```
Felixoid, line 106: Looks fine. Just an example of matching strings as a comment above would be nice

Felixoid (approved):

lacunoc (approved):

igcollect/linux_disk.py
konstantinbest: I had to change it again. It was also matching on xvda1 for example which we don't want.
The new regex `(xv|[shv]|nv)(d[a-z]|me[0-9]n[1-9])$` is matching to 
`xvda nvme0n1 vda sda`

PR #152, opened 2021-04-29
Include schema name for table sizes

elmatze79:
When multiple schemas use the same table names different values are returned for one name. This fix makes them distingushable.


axxelG (approved):
lgtm

PR #153, opened 2021-07-20
Optimize query to fetch relation sizes

dreis2211:
Hi,

rather than relying on `pg_statio_user_tables` that is doing way more than needed here, we just get the user table names
from `pg_class` and get their corresponding total relation size. By that we avoid some joins on `pg_index`, aggregation and function calls that we are not interested in here anyway.

Since `pg_total_relation_size(relid)` is acquiring an access share lock on the specific relation and therefore on all user tables, the query should be as fast as possible.

Scenario with quite some toast tables (600-700)
Before
```
| Execution Time: 21.253 ms
```
After
```
| Execution Time: 6.678 ms
```

### DIFF
--- a/igcollect/postgres.py
+++ b/igcollect/postgres.py
@@ -126,7 +126,7 @@ def main():
 
         # table size
         for line in execute(conn, ('''
-                SELECT relname, pg_total_relation_size(c.oid)
+                SELECT c.relname, pg_total_relation_size(c.oid)
                 FROM pg_class c
                 LEFT JOIN pg_namespace n ON n.oid = c.relnamespace
                 WHERE c.relkind IN ('r', 'm') AND n.nspname NOT IN ('pg_catalog', 'information_schema')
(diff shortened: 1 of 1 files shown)


igcollect/postgres.py
```diff
@@ -125,12 +125,13 @@ def main():
...
 
         # table size
-        for line in execute(conn, (
-                'SELECT relname,'
-                '    pg_total_relation_size(relid)'
-                "FROM pg_catalog.pg_statio_user_tables ORDER BY pg_total_relation_size(relid)"
-                "DESC;"
-        )):
+        for line in execute(conn, ('''
+                SELECT relname, pg_total_relation_size(relid)
```
kofrezo: This query does not work for me. There is no such column `relid` in `pg_class` table nor in `pg_namespace` table - am I holding it wrong ?

kofrezo (changes requested):

igcollect/postgres.py
dreis2211: Already fixed...Refresh

kofrezo (approved):

PR #154, opened 2021-07-29
Add support for more switch types

TuxPowered42:


igcollect/switch.py
```diff
@@ -223,6 +223,8 @@ def get_switch_model(snmp):
         return 'powerconnect'
     elif 'ProCurve' in model:
         return 'procurve'
+    elif 'Aruba' in model:
+        return 'procurve'
```
dreis2211, line 227: Might be a stupid question, but should that really return `procurve` in case of `Aruba`?

TuxPowered42: Nah, they are basically identical switches.

lamaral (approved):

PR #157, opened 2021-09-07
Standarize port name handling between igcollect and igmonplugins

TuxPowered42:


igcollect/switch.py
```diff
@@ -281,42 +292,15 @@ def get_laggs(snmp, model):
...
-    )
-    if g:
-        # Force 10 MXL port
-        return g.group(2).replace('/', '_')
-    if re.match('\Aethernet[0-9]+/[0-9]+\Z', port_name):
-        # Netiron ports
-        return port_name.replace('/', '_')
+    r = PORT_REGEXP[model].match(port_name)
+    if r:
+        g =  r.groupdict()['port']
```
lamaral, line 301: You can simplify this with `r.group('port')`

igcollect/switch.py
```diff
@@ -281,42 +292,15 @@ def get_laggs(snmp, model):
...
+        if g:
```
lamaral, line 302: This if doesn't make sense. 
If the group is not matched and you try to fetch it, you get an IndexError exception.

igcollect/switch.py
```diff
@@ -261,7 +272,7 @@ def get_monitored_ports(snmp, model):
     }
     for port_idx, port_state in port_states.items():
         if not laggs or port_idx not in laggs.keys():
-            port_name = standarize_portname(port_names[port_idx])
+            port_name = standarize_portname(port_names[port_idx], model)
```
lamaral, line 275: Typo: standardize

lamaral (changes requested):

igcollect/switch.py
```diff
@@ -75,6 +75,15 @@
     },
 }
 
+PORT_REGEXP = {
+    'cisco_ios': re.compile('^(?P<port>(Fa|Gi|Tu)[0-9/]+)$'),
+    'cumulus': re.compile('^(?P<port>swp[0-9]+)$'),
+    'extreme': re.compile('(?P<port>^[0-9]:[0-9]+)$'),
```
TheFRedFox, line 81:
```suggestion
    'extreme': re.compile('^(?P<port>[0-9]:[0-9]+)$'),
```

igcollect/switch.py
```diff
@@ -281,42 +292,15 @@ def get_laggs(snmp, model):
...
-        '\A(TenGigabitEthernet|fortyGigE) ([0-9]+/[0-9]+)\Z',
-        port_name
-    )
-    if g:
-        # Force 10 MXL port
-        return g.group(2).replace('/', '_')
-    if re.match('\Aethernet[0-9]+/[0-9]+\Z', port_name):
-        # Netiron ports
-        return port_name.replace('/', '_')
+    r = PORT_REGEXP[model].match(port_name)
```
TheFRedFox, line 299: What about a key that doesn't exist? KeyError being thrown is ok?

igcollect/switch.py
TheFRedFox: It's not just simplifying. It also makes the call None-safe. No KeyError will be raised for non existing keys.

igcollect/switch.py
TheFRedFox: It does again after changing `r.groupdict()['port']` to `r.group('port')`

TheFRedFox (approved):
Rather just trivial stuff. So, early approve.

igcollect/switch.py
lamaral: Right, I misunderstood the docs. After changing this to r.group(), it returns None instead of the Exception.

TuxPowered42: Fixed.

igcollect/switch.py
TuxPowered42: Fixed.

igcollect/switch.py
TuxPowered42: Fixed.

igcollect/switch.py
TuxPowered42: Fixed.

igcollect/switch.py
TuxPowered42: It's fine. We only support particular types of switches and the script crashing for any new switch is an indicator for us that SNMP OIDs for the new type must be researched and added.

igcollect/switch.py
```diff
@@ -281,44 +292,17 @@ def get_laggs(snmp, model):
...
-        return g.group(2).replace('/', '_')
-    if re.match('\Aethernet[0-9]+/[0-9]+\Z', port_name):
-        # Netiron ports
-        return port_name.replace('/', '_')
-
-    return None
+    r = PORT_REGEXP[model].match(port_name)
+    if not r:
+        return None
+    g =  r.group('port')
```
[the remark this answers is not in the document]
TuxPowered42: Fixed.

igcollect/switch.py
```diff
@@ -354,6 +338,11 @@ def cpu_stats(prefix, snmp, model):
         #     5 Secs ( 18.74%)    60 Secs ( 17.84%)   300 Secs ( 18.12%)
         m = re.search('60 Secs \( *([0-9]+)[0-9\.]*%\)', cpu_usage)
         cpu_usage = int(m.group(1))
+    elif model == 'cumulus':
+        # The value is percent idle
+        cpu_usage = 100 - cpu_usage
```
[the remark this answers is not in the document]
TuxPowered42: No, it depends on switch type.

lamaral (approved):

PR #160, opened 2022-02-02
Add sentry stats collector script

konstantinbest:
This script allows us to collect internal metrics (e.g. event processing queue length) from sentry and send them to graphite.

### DIFF
--- a/igcollect/sentry.py
+++ b/igcollect/sentry.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+"""Sentry stat collector
+
+Copyright (c) 2022 InnoGames GmbH
+"""
+from argparse import ArgumentParser
+from subprocess import run, PIPE, CalledProcessError
+from time import time
+
+
+def parse_args():
+    parser = ArgumentParser()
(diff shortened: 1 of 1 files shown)


igcollect/sentry.py
```diff
@@ -0,0 +1,35 @@
...
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument('--prefix', default='sentry.')
+    parser.add_argument('--config_path', default='/etc/sentry/sentry.conf.py')
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    proc = run(
```
seqizz, line 20: A small check if the run succeeded would be nice. In case the config is broken etc. At least to make script exit with a warning instead of crash.

igcollect/sentry.py
```diff
@@ -0,0 +1,35 @@
...
+        ["sentry", "--config", f"{args.config_path}", "queues", "list"],
```
kofrezo, line 21: Just a style issue: You can use single apostroh here too to be consistent with single apostroh with in the script.

seqizz (approved):

PR #161, opened 2022-03-14
linux_disklatency: pass path without libvirt + set size

brainexe:
small changes to linux_disklatency.py:
- new "path" option to directly set the target path for ioping (bypassing libvirt)
- add a new optional option to overwrite the default 4k size

Everything should be backward compatible with the old arguments

### DIFF
--- a/igcollect/linux_disklatency.py
+++ b/igcollect/linux_disklatency.py
@@ -7,27 +7,27 @@ Copyright (c) 2019 InnoGames GmbH
 from argparse import ArgumentParser
 from subprocess import check_output
 from time import time
-import sys
-from libvirt import (
-    open as libvirt_open,
-    libvirtError,
-)
 
 
 def parse_args():
     parser = ArgumentParser()
 
     parser.add_argument(
-        '--prefix', dest='prefix', type=str, default='None',
+        '--prefix', default='None',
         help='Graphite Metric Prefix')
 
     parser.add_argument(
-        '--storagepool', dest='storagepool', type=str, required=True,
-        help='Libvirt Storage Pool Name')
+        '--path', help='Path to check, by default libvirt is used.')
 
     parser.add_argument(
-        '--storagevol', dest='storagevol', type=str, required=True,
-        help='Libvirt Storage Volume Name')
+        '--size', default='4k',
+        help='Request size in byte')
+
+    parser.add_argument(
+        '--storagepool', help='Libvirt Storage Pool Name')
+
+    parser.add_argument(
+        '--storagevol', help='Libvirt Storage Volume Name')
 
     return parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


igcollect/linux_disklatency.py
```diff
@@ -22,11 +24,19 @@ def parse_args():
         help='Graphite Metric Prefix')
 
     parser.add_argument(
-        '--storagepool', dest='storagepool', type=str, required=True,
+        '--path', dest='path', type=str,
```
kofrezo, line 27: The `path=dest` is not necessary when it is the same as the variable name.

kofrezo: `type=str` is also not necessary as string is the  default.

igcollect/linux_disklatency.py
```diff
@@ -7,11 +7,13 @@
 from argparse import ArgumentParser
 from subprocess import check_output
 from time import time
-import sys
```
kofrezo, line 10: I think the `sys` module was present to allow using `sys.exit` instead of `exit`. All tough it wasn't used below we should use it.

See [this](https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python) StackOverflow post for a good explanation.

igcollect/linux_disklatency.py
```diff
@@ -22,11 +24,19 @@ def parse_args():
         help='Graphite Metric Prefix')
 
     parser.add_argument(
-        '--storagepool', dest='storagepool', type=str, required=True,
+        '--path', dest='path', type=str,
+        help='Path to check')
+
+    parser.add_argument(
+        '--size', dest='size', type=str,
+        help='Request size, default from ioping is 4k')
```
kofrezo, line 32: You could use `default` argument here. This makes it more readable and avoid hard coding the default int the comment.

Does it maybe make sense to use a integer here instead of string ?

igcollect/linux_disklatency.py
```diff
@@ -107,5 +87,45 @@ def main():
             print(template.format(mode, k, int(v)))
 
 
+def get_path(args) -> str:
+    if args.path is not None:
+        return args.path
+
+    # no --path is specified, use libvirt to get the storage volume
```
kofrezo, line 94: I think that would be helpful in the help of the `path` argument for the user.

igcollect/linux_disklatency.py
```diff
@@ -107,5 +87,45 @@ def main():
...
+    if args.storagepool is None or args.storagevol is None:
+        print('you have to set --storagepool AND --storagevol')
```
kofrezo, line 96: You could (try to) import the libvirt module only here if required. This would raise an exception if the module is not present which you want in this case because there would be no point of continueing in this case.

igcollect/linux_disklatency.py
```diff
@@ -107,5 +87,45 @@ def main():
...
+        exit(1)
```
kofrezo, line 97: Please use `sys.exit(1)` here (see previous comment at import for why).

kofrezo (approved):
Just some minor things to improve or discuss. Since the script would also work this way and non of the remarks would lead to a hard failure I am approving it already.

igcollect/linux_disklatency.py
```diff
@@ -107,5 +81,50 @@ def main():
...
+
+    # no --path is specified -> use libvirt to get the storage volume
+    from libvirt import (
+        open as libvirt_open,
+        libvirtError,
+    )
+
+    if args.storagepool is None or args.storagevol is None:
+        print('you have to set --storagepool AND --storagevol')
+        sys.exit(1)
```
TheFRedFox, line 96: `sys.exit`s should be just in the `main` function or it will get too confusing from where a script can be exited. Change it so the function is returning something or is raising an exception and the main function is handling it appropriately and doing the actual exit then.

igcollect/linux_disklatency.py
```diff
@@ -107,5 +81,50 @@ def main():
...
 
+def get_path(args) -> str:
+    if args.path is not None:
+        return args.path
+
+    # no --path is specified -> use libvirt to get the storage volume
+    from libvirt import (
+        open as libvirt_open,
+        libvirtError,
+    )
```
TheFRedFox, line 92: If I understood it correctly from before, it should not be clear and mandatory, that libvirt is available.
That would mean you would want to try and import those modules here:
``` python
try:
    from libvirt import ...
    # do libvirt stuff
except ImportError:
    # do alternative stuff

Comment (brainexe) [reply igcollect/linux_disklatency.py:88]: if no --path is set, "libvirt" is mandatory as it will extract the storage volumes based on the other parameters. Therefore no try/expect around it to get the plain exception in case (as before) .

Comment (TheFRedFox) [reply igcollect/linux_disklatency.py:88]: I see, then nevermind. You could try and check if the import fits in one line actually. Looks to me like it could. ^^

File: igcollect/linux_disklatency.py [hunk 9]
```diff
@@ -107,5 +77,46 @@ def main():
...
+
+    if args.storagepool is None or args.storagevol is None:
+        raise ArgumentError('if no --path is set you have to set --storagepool AND --storagevol')
```
TheFRedFox, line 91: It feels a bit weird to have script argument validation inside a helper function. But, I can live with that for now. ^^

igcollect/linux_disklatency.py
```diff
@@ -107,5 +77,46 @@ def main():
...
+
+    try:
+        conn = libvirt_open('qemu:///system')
+    except libvirtError as e:
+        raise Exception(
```
TheFRedFox, line 96: You can chain exceptions with `raise Exception(...) from e`. With that you still have the original exception in there.

TheFRedFox, line 96: As it is a small script, I think it is ok with using the broad `Exception` and just giving the note: Something more specific would be better. ;)

TheFRedFox (approved):

PR #163, opened 2022-04-06
jmx: Make metric-file configurable

mpeltekis:


igcollect/jmx.py
```diff
@@ -19,6 +19,7 @@ def parse_args():
     parser.add_argument('--ports', default=[], nargs='+')   # XXX Deprecated
     parser.add_argument('--names', default=[], nargs='+')   # XXX Deprecated
     parser.add_argument('--thread-prefixes', default=[], nargs='+')
+    parser.add_argument('--metric-file', default='etc/jmx2graphite/metrics', nargs='?')
```
kofrezo, line 22: Specifying nargs with `?` will also allow passing the argument without value e.g. `--metric-file` instead of `--metric-file /etc/something`.

You may want to specify const or use a different nargs value. You can consult the [docs](https://docs.python.org/3/library/argparse.html?highlight=argument%20parser#nargs)

igcollect/jmx.py
```diff
@@ -59,6 +60,8 @@ def main():
                 port,
                 '--thread-prefixes',
                 thread_prefixes,
+                '--metric-file',
+                args.metric_file,
```
kofrezo, line 64: This will also add the argument to all current users. Is that fine (is it the implicit default) ?

kofrezo (commented):
Please double check the comments from me and @TheFRedFox. 

If fine I have no objections.

igcollect/jmx.py
mpeltekis: to my understanding is passing it without argument will always produce the default which is correct here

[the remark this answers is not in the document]
mpeltekis: yes changed it

mpeltekis: if its emtpy it now should now throw an error jmx.py: error: argument --metric-file: expected one argument

PR #164, opened 2022-05-02
Add EPYC 7763 per factor

raimund-s:
Add a rather conservative perf factor to compensate for the higher memory pressure compared to a 32 core Epyc


igcollect/linux_cpu_perffactor.py
```diff
@@ -27,7 +27,8 @@ def main():
         'Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz': 1.75,
         'Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz': 1.8,
         'Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz': 1.875,
-        'AMD EPYC 7502P 32-Core Processor': 2.25
+        'AMD EPYC 7502P 32-Core Processor': 2.25,
+        'AMD EPYC 7763 64-Core Processor': 2.25
```
TuxPowered42, line 31: Missing `,`.

raimund-s: added

PR #167, opened 2022-05-25
Allow parsing of timestamps with high precision

meikomeik:


meikomeik:
This MR fixes the bug like:
`ValueError: invalid literal for int() with base 10: '2022-05-25T11:54:01.425646721+0000'`

igcollect/logfile_values.py
```diff
@@ -256,6 +256,16 @@ def convert_to_timestamp(time_str, time_format):
...
         as local
     """
+    # When using Java17 we are seeing higher decimal fractions on the dates
+    # that Python isn't able to handle.   We need to make sure that the
+    # decimal fraction part is only 6 digits long.
+    # The regexp replaces a 9 digit number with a 6 digit one
+    # time_str example: '2022-05-25T12:05:15.654320355Z'
+    _time_str = time_str.split('.', 1)
+    if len(_time_str[1]) > 7:
+        _time_str[1] = re.sub(r'^([0-9]{6})[0-9]{3}(Z)?', r'\1\2', _time_str[1])
```
TheFRedFox, line 266: Another way could be to split on `Z` again and just cut off everything after 6 characters:
```suggestion
        _tmp, _ = _time_str[1].split('Z', 1)
        _time_str[1] = _tmp[:6] + 'Z'
```

meikomeik: is a regexp such a bad idea?

TheFRedFox, line 266: If with regexp, I am wondering, if it would make more sense to make the second number sequence infinite:
```suggestion
        _time_str[1] = re.sub(r'^([0-9]{6})[0-9]+(Z)?', r'\1\2', _time_str[1])
```
or
```suggestion
        _time_str[1] = re.sub(r'^([0-9]{6})[0-9]*(Z)?', r'\1\2', _time_str[1])
```
Advantage would be to not touch it again if it gets even more precise.

meikomeik: Would be fine with me

TheFRedFox: No, not at all, just that it uses more resources and I had an idea to do the same without regexp, which would save some CPU and memory performance.

meikomeik: Changed it to a +

TheFRedFox (approved):
Just smaller discussions. Early approve.

PR #170, opened 2022-08-05
Add SFP DOM metric collection script

lamaral:


igcollect/sfp_dom.py
```diff
@@ -0,0 +1,372 @@
...
+    # Get module info
+    em = ethtool_modinfo()
+    em.cmd = ETHTOOL_GMODULEINFO
+    ifr.ifr_data.ethtool_modinfo_ptr = ctypes.pointer(em)
+
+    try:
+        fcntl.ioctl(sock, SIOCETHTOOL, ifr)
+    except OSError as exc:
+        if exc.errno == 95:
+            raise NoSFPException(f'Interface {interface} has no SFP module')
```
Woellchen, line 336: maybe you want to `raise .. from exc`

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,372 @@
...
+    ee = ethtool_eeprom()
+    ee.cmd = ETHTOOL_GMODULEEEPROM
+    ee.len = em.eeprom_len
+    ee.offset = 0
+    ifr.ifr_data.ethtool_eeprom_ptr = ctypes.pointer(ee)
+
+    try:
+        fcntl.ioctl(sock, SIOCETHTOOL, ifr)
+    except OSError as exc:
+        raise
```
Woellchen, line 349: it looks like this `try..except` can be removed

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,372 @@
...
+
+    if em.type == ETH_MODULE_SFF_8079:
```
Woellchen, line 351: you could refactor these elifs to something like `if em.type in (ETH_MODULE_SFF_8079, ...)`

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,372 @@
...
+
+    return physical_interfaces
+
+
+def a2_offset_to_celsius(data, offset):
+    """
+    Returns a signed float built from a 16-bit signed twos complement value
+    in increments of 1/256 degrees Celsius
+    """
+    signed = ctypes.c_int16((data[SFF_A2_BASE + (offset)] << 8 | data[
```
Woellchen, line 289: you can call `a2_offset_to_int()` here

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,372 @@
...
+        ('ifr_name', (ctypes.c_ubyte * IFNAMSIZ)),
+        ('ifr_data', ifr_data)
+    ]
+
+
+class SFPData:
+    def __init__(self, data, interface):
+        self._data = data
+        self.interface = interface
+        self.metrics = {
```
Woellchen, line 111: you could make this a constant on the top, something like this:
```
METRICS_MAP = {
    'temperature': {
        'current': SFF_A2_TEMP,
        'low_warning': SFF_A2_TEMP_LWARN,
        ...
    },
    'bias_current': {
        'current': SFF_A2_BIAS,
        ...
    },
    ...
}
```
also you can define another fn map like so:
```
FN_MAP = {
    'temperature': a2_offset_to_celsius,
    ...
}
```

that way you can do the following:
- initialize your empty metrics dict here with dict comprehension
- map the right conversion functions to the right data points

however you would need to slightly rearrange some stuff so you can actually reference it

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,372 @@
...
+            self._data, SFF_A2_TX_PWR_LALRM
+        )
+        self.metrics['tx_power']['high_critical'] = a2_offset_to_volts(
+            self._data, SFF_A2_TX_PWR_HALRM
+        )
+
+        # RX Power is 16-bit unsigned integer with LSB equal to 0.1uW
+        # We use the a2_offset_to_volts() function because the equivalence
+        # is similar to volts
+        rx_power = a2_offset_to_int(self._data, SFF_A2_RX_PWR) / 10000
```
Woellchen, line 232: is `rx_power` a leftover?

Woellchen (changes requested):

igcollect/sfp_dom.py
lamaral: Yes, it was a left over.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,364 @@
...
+SFF_A2_TX_PWR = 102
+SFF_A2_TX_PWR_HALRM = 24
+SFF_A2_TX_PWR_LALRM = 26
+SFF_A2_TX_PWR_HWARN = 28
+SFF_A2_TX_PWR_LWARN = 30
+SFF_A2_RX_PWR = 104
+SFF_A2_RX_PWR_HALRM = 32
+SFF_A2_RX_PWR_LALRM = 34
+SFF_A2_RX_PWR_HWARN = 36
+SFF_A2_RX_PWR_LWARN = 38
```
TuxPowered42, line 71: Please document where those values are coming from. The names of the .h files will be enough.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,364 @@
...
+            self._data, SFF_A2_TEMP_HWARN
+        )
+        self.metrics['temperature']['low_critical'] = a2_offset_to_celsius(
+            self._data, SFF_A2_TEMP_LALRM
+        )
+        self.metrics['temperature']['high_critical'] = a2_offset_to_celsius(
+            self._data, SFF_A2_TEMP_HALRM
+        )
+
+        # Bias current is 16-bit unsigned integer with LSB equal to 2uA
```
TuxPowered42, line 176: That's `µ`, not `u`.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,364 @@
...
+            self._data, SFF_A2_BIAS_HWARN
+        )
+        self.metrics['bias_current']['low_critical'] = a2_offset_to_milliamps(
+            self._data, SFF_A2_BIAS_LALRM
+        )
+        self.metrics['bias_current']['high_critical'] = a2_offset_to_milliamps(
+            self._data, SFF_A2_BIAS_HALRM
+        )
+
+        # Voltage is 16-bit unsigned integer with LSB equal to 100uV
```
TuxPowered42, line 193: That's `µ`, not `u`.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,364 @@
...
+            self._data, SFF_A2_VCC_HWARN
+        )
+        self.metrics['sfp_voltage']['low_critical'] = a2_offset_to_volts(
+            self._data, SFF_A2_VCC_LALRM
+        )
+        self.metrics['sfp_voltage']['high_critical'] = a2_offset_to_volts(
+            self._data, SFF_A2_VCC_HALRM
+        )
+
+        # TX Power is 16-bit unsigned integer with LSB equal to 0.1uW
```
TuxPowered42, line 210: That's `µ`, not `u`.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,364 @@
...
+            self._data, SFF_A2_TX_PWR_HWARN
+        )
+        self.metrics['tx_power']['low_critical'] = a2_offset_to_volts(
+            self._data, SFF_A2_TX_PWR_LALRM
+        )
+        self.metrics['tx_power']['high_critical'] = a2_offset_to_volts(
+            self._data, SFF_A2_TX_PWR_HALRM
+        )
+
+        # RX Power is 16-bit unsigned integer with LSB equal to 0.1uW
```
TuxPowered42, line 229: That's `µ`, not `u`.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,364 @@
...
+    signed = ctypes.c_int16(a2_offset_to_int(data, offset))
+    return signed.value / 256
+
+
+def a2_offset_to_volts(data, offset):
+    return a2_offset_to_int(data, offset) / 10000
+
+
+def a2_offset_to_milliamps(data, offset):
+    return a2_offset_to_int(data, offset) / 256
```
TuxPowered42: The conversion in line 292 is done by a decimal number 10k to convert a 100µV scale into V. 100µ * 10k = 1M and 1/1M is 1µ.  But here for the current which is given in 1µA units you divide by 256. Shouldn't you divide by 1k?

TuxPowered42 (changes requested):

igcollect/sfp_dom.py
lamaral: I ended up changing this to a struct.unpack.

igcollect/sfp_dom.py
lamaral: Done.

igcollect/sfp_dom.py
lamaral: I caught this mistake. 
The division here should have been 500, as each LSB is equivalent to 2µA.

igcollect/sfp_dom.py
lamaral: Changed.

igcollect/sfp_dom.py
lamaral: I decided to include the maps in the SFPData class. It made more sense to put them there, especially because the FN_MAP needed the functions to be declared beforehand.

lamaral:
I think I addressed all the requested changes. 
A second look is appreciated :)

igcollect/sfp_dom.py
TheFRedFox: I second this. Was already thinking about that while reading through the code. The different calculation functions were already screaming for that in my mind. ^^ It would go in the direction of SOLID-O (so, open for extension / closed for modification), which would mean that `SFPData` or at least the method `SFPData::_parse_eeprom_data` would be open for extension while being closed for modification. Meaning, that to add more functionality, you could do that then without touching the function (or class) at all. This would make it easier and faster to maintain.

Maybe another idea could be to not have 2 configuration points:
``` python
METRICS_MAP = {
    'temperature': {
        'current': (SFF_A2_TEMP, a2_offset_to_celsius),
        'low_warning': (SFF_A2_TEMP_LWARN, a2_offset_to_celsius),
        ...
    },
    'bias_current': {
        'current': (SFF_A2_BIAS, a2_offset_to_milliamps),
        ...
    },
    ...
}
```

You could use that then like that:
``` python
def _parse_eeprom_data(self):
    ...
    for category, sub_metrics in METRICS_MAP.items():
        for metric, (offset, fn) in sub_metrics.items():
            self.metrics[category][metric] = fn(self._data, offset)
```

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+        self.metrics['rx_power']['low_critical'] = a2_offset_to_volts(
+            self._data, SFF_A2_RX_PWR_LALRM
+        )
+        self.metrics['rx_power']['high_critical'] = a2_offset_to_volts(
+            self._data, SFF_A2_RX_PWR_HALRM
+        )
+
+
+class NoSFPException(Exception):
```
TheFRedFox, line 340: Good thing that you use own Exceptions! 👍🏼

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+        for value_key, value in child.items():
+            print(
+                f'{prefix}.{interface}.{metric}.{value_key}'
+                f' {value} {timestamp}'
+            )
+
+
+def ethtool_get_module_data(interface):
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    name = (ctypes.c_ubyte * IFNAMSIZ)(*bytearray(str(interface).encode()))
```
TheFRedFox, line 167: I think that could have some comment as that looks like it is doing quite some magic. ^^

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+    ifr = ifreq()
+    ifr.ifr_name = name
```
TheFRedFox, line 169: Is it maybe possible to make the constructor of `ifreq` so it take the name already in it?
``` python
ifr = ifreq(ifr_name=name)
```

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+
+    # Get module info
+    em = ethtool_modinfo()
+    em.cmd = ETHTOOL_GMODULEINFO
```
TheFRedFox, line 173: Is it maybe possible to make the constructor of `ethtool_modinfo` so it take the name already in it?
``` python
em = ethtool_modinfo(cmd=ETHTOOL_GMODULEINFO)
```

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+    ifr.ifr_data.ethtool_modinfo_ptr = ctypes.pointer(em)
```
TheFRedFox, line 174: Is it maybe possible to make the constructor of `ifreq` so it take the name already in it?
``` python
ifr = ifreq(ethtool_ptr=ctypes.pointer(em))
```

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+        if exc.errno == 95:
+            raise NoSFPException(f'Interface {interface} has no SFP module') \
+                from exc
+        else:
+            raise
+
+    ee = ethtool_eeprom()
+    ee.cmd = ETHTOOL_GMODULEEEPROM
+    ee.len = em.eeprom_len
+    ee.offset = 0
```
TheFRedFox, line 188: Is it maybe possible to make the constructor of `ethtool_eeprom` so it take the name already in it?
``` python
ifr = ethtool_eeprom(cmd=ETHTOOL_GMODULEEEPROM, len=eeprom_len, offset=0)
```

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+def main():
+    args = parse_args()
+    physical_interfaces = get_physical_interfaces()
+
+    for iface in physical_interfaces:
+        try:
+            data = ethtool_get_module_data(iface)
+            print_metrics(args.prefix, iface, data.metrics)
```
TheFRedFox, line 93: `ethtool_get_module_data` could also return `None`. Is it ok that it will throw a `NoneType` exception while trying to read `data.metrics`?

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+            print(exc, file=sys.stderr)
+            continue
+
+
+def get_physical_interfaces():
+    scn = '/sys/class/net'
+    physical_interfaces = []
+    for dev in listdir(scn):
+        dev_path = join(scn, dev)
+        if not (islink(dev_path) or isdir(dev_path)):
```
TheFRedFox, line 104: You are checking extra for `isdir`, but it seems that you are not doing anything afterwards with that information, as you are again checking for `islink` again.
Is that correct?

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+Copyright (c) 2022 InnoGames GmbH
+
+This script reads SFP EEPROM and parses it in order to collect optical
+diagnostics data.
+
+To properly understand what happens on this script, a read of the SFF-8472
+standard is highly recommended.
+
+In some places, we gave up on making things Pythonic in detriment of shaving
+off a few ms on execution time.
```
TheFRedFox, line 13: Is this super crucial? Then I will check a bit more over it.

igcollect/sfp_dom.py
TheFRedFox: This might be a bit slower though than the direct usage of the functions.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+                'high_warning': None,
+                'low_critical': None,
+                'high_critical': None,
+            },
+        }
+        self._parse_eeprom_data()
+
+    def _parse_eeprom_data(self):
+        # Check if SFP data is externally calibrated
+        self.external_calibration = self._data[SFF_A0_DOM] & SFF_A0_DOM_EXTCAL
```
TheFRedFox, line 327: You are never using that again. So, I doubt that it is needed to store it on `self`.

Besides, I would also suggest to do that directly during `__init__` or can `self._data` change between `__init__` and `_parse_eeprom_data` execution?

TheFRedFox (commented):
Several discussion points.

igcollect/sfp_dom.py
```diff
@@ -0,0 +1,371 @@
...
+            ETH_MODULE_SFF_8079, ETH_MODULE_SFF_8636, ETH_MODULE_SFF_8436
+    ):
+        raise UnsupportedSFPException(f'SFP type {em.type} is not supported')
+    elif em.type == ETH_MODULE_SFF_8472:
+        return SFPData(ee.data, interface)
+
+    return None
+
+
+class ethtool_modinfo(ctypes.Structure):
```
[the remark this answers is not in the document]
lamaral: Actually, nothing speaks against it. I will adjust them accordingly.

igcollect/sfp_dom.py
lamaral: Actually, not. I did some rudimentary benchmarking and the difference was negligible, so I removed them. I will remove this part.

igcollect/sfp_dom.py
lamaral: The functions differ per the outer key of the `METRICS_MAP` dict. 
I feel like adding the function on each inner key is too redundant. 

I thought about:
```python
METRICS_MAP = {
    'temperature': {
        'function': a2_offset_to_celsius,
        'current': SFF_A2_TEMP,
        'low_warning': SFF_A2_TEMP_LWARN,
        ...
    },
    'bias_current': {
        'function': a2_offset_to_milliamps,
        'current': SFF_A2_BIAS,
        ...
    },
    ...
}
Since we iterate over `self.metrics`, it should be fine key-wise.

File: igcollect/sfp_dom.py (hunk 14 above)
Comment (lamaral) [reply igcollect/sfp_dom.py:174]: I added some more comments for this function on 4f533f4

File: igcollect/sfp_dom.py (hunk 15 above)
Comment (lamaral) [reply igcollect/sfp_dom.py]: I adjusted the code to match the example above.

File: igcollect/sfp_dom.py (hunk 16 above)
Comment (lamaral) [reply igcollect/sfp_dom.py]: I adjusted the code to match the example above.

File: igcollect/sfp_dom.py (hunk 17 above)
Comment (lamaral) [reply igcollect/sfp_dom.py:181]: I think this simplification won't work here, because `ifr.ifr_data` is a pointer to an union. I would have to instantiate an IFRData object separately as well, which I think complicates things way more.

Please keep in mind that these are ctypes, not native Python structures.

File: igcollect/sfp_dom.py (hunk 18 above)
Comment (lamaral) [reply igcollect/sfp_dom.py]: I adjusted the code to match the example above.

File: igcollect/sfp_dom.py (hunk 22 above)
Comment (lamaral) [reply igcollect/sfp_dom.py]: It makes sense. 
I moved the check up and removed the `self`.

File: igcollect/sfp_dom.py (hunk 20 above)
Comment (lamaral) [reply igcollect/sfp_dom.py:101]: I took this from another igcollect script and didn't realize we didn't need this first if. 
I have simplified the code here.

File: igcollect/sfp_dom.py (hunk 19 above)
Comment (lamaral) [reply igcollect/sfp_dom.py:90]: I refactored the code in `ethtool_get_module_data` a little bit to prevent `None` from being returned.

**Comment by lamaral** [comment]
All review change requests are implemented.

**Review by Woellchen — APPROVED** [review]
awesome!!

File: igcollect/sfp_dom.py [hunk 24]
```diff
@@ -0,0 +1,353 @@
...
+                'high_critical': None,
+            },
+            'rx_power_lane1': {
+                'current': None,
+                'low_warning': None,
+                'high_warning': None,
+                'low_critical': None,
+                'high_critical': None,
+            },
+        }
```
TheFRedFox, line 333: To reduce some points to touch when the mappings changes, I would make this here dynamic:
```suggestion
        self.metrics = {cat: {metric: None for metric, _ in m_offs.items()} for cat, m_offs in OFFSET_MAP.items()}
```

TheFRedFox (approved):
I think one more thing would be beneficial. The rest looks good now. :) So, early approve.

igcollect/sfp_dom.py
TheFRedFox: I would actually prefer to have just one point of configuration. So, we won't have duplicated configuration points. This would make it easier to maintain.

Your suggestion would be a bit tricky, as the you put the `function` on the same level as the `metric` names. To avoid that, I would suggest to either:
``` python
METRICS_MAP = {
    'temperature': {
        '__function': a2_offset_to_celsius,
        'current': SFF_A2_TEMP,
        'low_warning': SFF_A2_TEMP_LWARN,
        ...
    },
    'bias_current': {
        '__function': a2_offset_to_milliamps,
        'current': SFF_A2_BIAS,
        ...
    },
    ...
}
```

or

``` python
METRICS_MAP = {
    'temperature': {
        'function': a2_offset_to_celsius,
        'metrics': {
            'current': SFF_A2_TEMP,
            'low_warning': SFF_A2_TEMP_LWARN,
            ...
        }
    },
    'bias_current': {
        'function': a2_offset_to_milliamps,
        'metrics': {
            'current': SFF_A2_BIAS,
            ...
        }
    },
    ...
}
```

lamaral: I will send a second PR soon extending some of the functionality and adding a new type of data structure that will address this.
There were quite some changes around this part, so I'd rather leave it as is for this PR. I took these comments into consideration when writing the new set of patches.

igcollect/sfp_dom.py
lamaral: Same as above. I will address this in the new PR that I will submit.

TuxPowered42 (approved):

PR #171, opened 2022-08-10
Collect SFP DOM Metrics for MXL10/40 switches

lamaral:


igcollect/switch.py
```diff
@@ -328,6 +355,23 @@ def ports_stats(prefix, snmp, ports, model):
...
 
 
+def dom_stats(prefix, snmp, ports, model):
+    """
+    Print graphite-compatible Digital Optical Monitoring stats for each port
+    of the switch
+    """
+
+    for metric, oid in DOM_OIDS[model].items():
+        # SNMP is slow, so we want need to get current time for each counter.
```
TuxPowered42, line 365: If it's so slow, then we should change this behaviour for all OIDs. Please check how it changes the total execution time and CPU load.

igcollect/switch.py
```diff
@@ -52,6 +53,25 @@
...
+        'rx_power.0': '1.3.6.1.4.1.6027.3.11.1.3.1.1.12',
+        'rx_power.1': '1.3.6.1.4.1.6027.3.11.1.3.1.1.13',
+        'rx_power.2': '1.3.6.1.4.1.6027.3.11.1.3.1.1.14',
+        'rx_power.3': '1.3.6.1.4.1.6027.3.11.1.3.1.1.15',
+        'sfp_voltage': '1.3.6.1.4.1.6027.3.11.1.3.1.1.17',
+        'temperature': '1.3.6.1.4.1.6027.3.11.1.3.1.1.16',
+        'tx_power.0': '1.3.6.1.4.1.6027.3.11.1.3.1.1.8',
+        'tx_power.1': '1.3.6.1.4.1.6027.3.11.1.3.1.1.9',
+        'tx_power.2': '1.3.6.1.4.1.6027.3.11.1.3.1.1.10',
+        'tx_power.3': '1.3.6.1.4.1.6027.3.11.1.3.1.1.11',
```
TuxPowered42, line 71: What happens on port with just a single transceiver? I don't think we have any 100Gb/s ports on Dell switches.

TuxPowered42 (changes requested):

igcollect/switch.py
lamaral: We are not talking about transceivers, but about lanes on the transceivers. 

Ports with transceivers with a single lane will output an empty value for the other lanes that gets ignored on line 371:
```python
                if data and data != '':
                    print(template.format(port_name, metric, data))
```

igcollect/switch.py
lamaral: Same comment as line 342. It was already there. 

I will check the impact on these metrics, nonetheless.

igcollect/switch.py
```diff
@@ -115,6 +137,11 @@ def main():
     monitored_ports = get_monitored_ports(snmp, model)
     ports_stats(args.prefix, snmp, monitored_ports, model)
 
+    # We check DOM metrics only for switch models that have OIDs added to
+    # the script
+    if model in DOM_OIDS:
+        dom_stats(args.prefix, snmp, monitored_ports, model)
```
Woellchen, line 143: this is a minor, but you could already pass `DOM_OIDS[model]` into the function to decouple it from this scope

igcollect/switch.py
```diff
@@ -328,6 +355,23 @@ def ports_stats(prefix, snmp, ports, model):
...
+    """
+
+    for metric, oid in DOM_OIDS[model].items():
+        # SNMP is slow, so we want need to get current time for each counter.
+        template = prefix + '.ports.{}.{} {} ' + str(int(time()))
+        table = get_snmp_table(snmp, oid)
+        for port_idx, port_name in ports.items():
+            if port_idx in table:
+                data = table[port_idx]
+                if data and data != '':
```
Woellchen, line 371: i like to keep indentations as few as possible. you could negate the ifs and `continue` early if you like

igcollect/switch.py
```diff
@@ -328,6 +355,23 @@ def ports_stats(prefix, snmp, ports, model):
...
+                    print(template.format(port_name, metric, data))
```
TheFRedFox, line 372: Suggestion to @Woellchen 's comment:
```suggestion
        for port_idx, port_name in ports.items():
            if port_idx not in table:
                continue
            data = table[port_idx]
            if not data:
                continue
            print(template.format(port_name, metric, data))
```

igcollect/switch.py
TheFRedFox, line 365: `.., so we want need ...`, want or need?

igcollect/switch.py
```diff
@@ -328,6 +355,23 @@ def ports_stats(prefix, snmp, ports, model):
                 print(template.format(port_name, counter, data))
 
 
+def dom_stats(prefix, snmp, ports, model):
+    """
+    Print graphite-compatible Digital Optical Monitoring stats for each port
+    of the switch
+    """
+
+    for metric, oid in DOM_OIDS[model].items():
```
TheFRedFox, line 364: From the scope of the function it can happen that `DOM_OIDS[model]` will result in a KeyError.

You could avoid it altogether with @Woellchen 's suggestion in his [comment](https://github.com/innogames/igcollect/pull/171/files/1ab699a4271f8e35fd7d02b664e08a1023a75ed4#r944546214).
Or thinking about what should happen if a model is not in `DOM_OIDS`.

TuxPowered42 (approved):

igcollect/switch.py
lamaral: Good idea, I adjusted that.

igcollect/switch.py
lamaral: I adjusted it like Freddy's suggestion.

igcollect/switch.py
lamaral: Thank you. I committed the suggestion.

igcollect/switch.py
lamaral: That's fixed.

igcollect/switch.py
lamaral: I went with Patrick's suggestion.

Woellchen (approved):

igcollect/switch.py
lamaral: This specific patch for MXL switches raised the execution time from 3,5 to 5,5 seconds. 
The user and system times kept pretty much the same, as most of the time is spent waiting for the SNMP responses.

PR #172, opened 2022-08-17
Add support for QSFP (SFF8636) metric collection

lamaral:


igcollect/sfp_dom.py
```diff
@@ -105,54 +152,76 @@ def get_physical_interfaces():
 
 
 def a2_offset_to_celsius(data, offset):
+    return offset_to_celsius(data, SFF_A2_BASE + offset)
+
+
+def a2_offset_to_volts(data, offset):
+    return offset_to_int(data, SFF_A2_BASE + offset) / 10000
```
Woellchen, line 159: you could call `offset_to_volts()` here, no?

igcollect/sfp_dom.py
```diff
@@ -105,54 +152,76 @@ def get_physical_interfaces():
...
+
+
+def a2_offset_to_milliwatts(data, offset):
+    return offset_to_int(data, SFF_A2_BASE + offset) / 10000
```
Woellchen, line 163: you could call `offset_to_milliwatts()` here, no?

igcollect/sfp_dom.py
```diff
@@ -105,54 +152,76 @@ def get_physical_interfaces():
...
+
+
+def a2_offset_to_milliamps(data, offset):
+    return offset_to_int(data, SFF_A2_BASE + offset) / 500
```
Woellchen, line 167: you could call `offset_to_milliamps()` here, no?

igcollect/sfp_dom.py
```diff
@@ -235,51 +306,161 @@ class IFReq(ctypes.Structure):
...
-        'bias_current_lane1': a2_offset_to_milliamps,
-        'sfp_voltage': a2_offset_to_volts,
-        'tx_power_lane1': a2_offset_to_milliwatts,
-        'rx_power_lane1': a2_offset_to_milliwatts,
+class SFFChannelData:
+    bias_current = None
+    rx_power = None
+    tx_power = None
+
+    def __init__(self):
```
Woellchen, line 314: you can remove the empty `__init__`

igcollect/sfp_dom.py
```diff
@@ -235,51 +306,161 @@ class IFReq(ctypes.Structure):
...
 
-class SFPData:
-    FN_MAP = {
-        'temperature': a2_offset_to_celsius,
-        'bias_current_lane1': a2_offset_to_milliamps,
-        'sfp_voltage': a2_offset_to_volts,
-        'tx_power_lane1': a2_offset_to_milliwatts,
-        'rx_power_lane1': a2_offset_to_milliwatts,
+class SFFChannelData:
+    bias_current = None
```
Woellchen, line 310: maybe you want to use dataclasses here, or old-school `@property`. then you don't need the `values` getter and setter

igcollect/sfp_dom.py
lamaral: Fixed.

igcollect/sfp_dom.py
lamaral: Fixed.

igcollect/sfp_dom.py
lamaral: Fixed.

igcollect/sfp_dom.py
lamaral: Fixed

igcollect/sfp_dom.py
lamaral: I gave dataclasses a try. 

The biggest issue is because I pre-create the `SFFChannelData` objects when I create the `SFFData` object, but I only set the values later, when `SFFData` calls the `_parse_eeprom_data()` function. 

I also rely later on the data from `SFFChannelData` already in a dict, which is provided by the `get_values()` function.

igcollect/sfp_dom.py
```diff
@@ -251,26 +257,35 @@ def ethtool_get_module_data(interface):
...
         else:
             raise
 
     # Build the struct with the ETHTOOL_GMODULEEEPROM command and request.
     # Objects below are ctypes structs that will be later used with the
     # ioctl calls.
-    ee = EthtoolEEPROM(cmd=ETHTOOL_GMODULEEEPROM, len=em.eeprom_len, offset=0)
+    # We limit the length to 604, as that is the biggest size of an EEPROM
+    # according to the standard and we don't want buffer overruns.
+    length = 604 if em.eeprom_len > 604 else em.eeprom_len
```
Woellchen, line 273: `min(604, em.eeprom_len)` would be a bit more elegant, and while at it make `604` a constant maybe

lamaral: Both done :)

Woellchen (approved):

PR #174, opened 2022-09-05
Always return the current as well

raimund-s:
Artfiles changed the behaviour that now the current current is returned for every entry also if the PDU is billed by kWh.
Also add the 32A option and remove python 2 compatibility import


TheFRedFox (approved):
Can you just fix the typo in your second commit? ^^

PR #179, opened 2023-02-08
Add a collector for netfilter stats

TuxPowered42:


igcollect/netfilter.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def get_snat_info(snats):
+    ret = {}
+    for snat in snats:
+        snat_states = 0
+        for line in check_output(
+            ['/sbin/conntrack', '-L', '-n', '-q', snat],
+        ).decode().splitlines():
+            snat_states += 1
```
lamaral, line 42: Why a for? You cloud just call `len()` and assign that so snat_states.

lamaral (changes requested):

igcollect/netfilter.py
```diff
@@ -0,0 +1,81 @@
...
+    nft = Nftables()
+    nft.set_json_output(True)
+    _, nft_output, _ = nft.cmd(f"list counters")
+
+    for data in json.loads(nft_output)['nftables']:
+        counter = data.get('counter')
+        if not counter:
+            continue
+        ret[f'counters.{counter["name"]}.packets'] = counter['packets']
+        ret[f'counters.{counter["name"]}.bytes'] = counter['bytes']
```
lamaral, line 76: Should we maybe ensure a safe_name for the counters here, just in case?

lamaral (approved):

PR #185, opened 2024-02-13
Draft: Collect UMA stats

TuxPowered42:


igcollect/freebsd_memory.py
```diff
@@ -1,21 +1,71 @@
...
+    for line in sysctl.filter('vm.uma'):
+        s = line.name.split('.')
+        if len(s) == 3:
+            uma_info[s[2]] = {}
+
+    # Fetch the real data in a separate step only for wanted OIDs.
+    for oid in uma_info.keys():
+        for metric in UMA_STATS:
+            for line in sysctl.filter(f'vm.uma.{oid}.{metric}'):
+                uma_info[oid][metric] = parse_sysctl_value(line)
```
lamaral, line 64: Do you really need this for loop? 
The left side of the assignment doesn't change with the for, so you either have a single element or you are overwriting data.

TuxPowered42:
Replaced with newer code

PR #186, opened 2024-06-10
Add igcollect script for Chrony

lamaral:


igcollect/chrony.py
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    for category, data in metrics.items():
+        for metric, value in data.items():
+            print(f'{args.prefix}.{category}.{metric} {value} {now}')
+
+
+def run_chronyc_command(command):
+    chrony_info_raw = check_output(
+        ['/usr/bin/chronyc', '-c', command],
```
TuxPowered42, line 35: Will this check ever run on FreeBSD? Chrony is not from base system but from a package. The path is `/usr/local/bin/chronyc`.

TuxPowered42 (changes requested):

igcollect/chrony.py
lamaral: I forgot to account for it. I pushed a fix that checks the platform.

TuxPowered42 (approved):

PR #192, opened 2025-06-03
postgres: add lock_timeout to avoid ping up queries in case of longer locks

brainexe:
We had the case of longer exclusive locks on a table. This caused pining up monitoring queries every minute, hanging and blocking the recovering from primary.


kofrezo:
> We had the case of longer exclusive locks on a table. This caused pining up monitoring queries every minute, hanging and blocking the recovering from primary.

Did you figure out what is causing the locks here ? We are doing only select queries ...

kofrezo (approved):

PR #200, opened 2025-09-10
Fix logfile_values.py to support key-value log formats

konstantinbest:
- Add support for parsing key-value pairs (e.g., status=200, timestamp="2025-09-10T08:01:21+00:00")
- Maintain full backward compatibility with existing plain column log formats
- Enable parsing of modern structured log formats like nginx key-value logs

This fixes issues with nginx logs that have changed from space-separated columns to key-value format, allowing status code distribution analysis to work correctly.

### DIFF
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -71,21 +71,11 @@ class Metric:
 
         Returns 0.0 if the value cannot be converted to float.
         """
-        # If it's a key-value pair, extract the value
-        if '=' in field:
-            key, value = field.split('=', 1)
-            # Remove quotes if present
-            value = value.strip('"')
-            try:
-                return float(value)
-            except ValueError:
-                return 0.0
-        else:
-            # Plain value
-            try:
-                return float(field)
-            except ValueError:
-                return 0.0
+        raw_value = extract_value(field)
+        try:
+            return float(raw_value)
+        except ValueError:
+            return 0.0
 
     def estimate_columns_value(self, fields):
         """
@@ -211,7 +201,13 @@ def parse_args():
 
 def extract_value(field):
     """
-    Extract value from key-value pair or plain value
+    Extract raw value from key-value pair or plain value
+    
+    Args:
+        field: String field that may be in format "key=value" or just "value"
+        
+    Returns:
+        The extracted value as string (with quotes stripped if present)
     """
     if '=' in field:
         key, value = field.split('=', 1)
--- a/igcollect/logfile_values.py
+++ b/igcollect/logfile_values.py
@@ -65,11 +65,41 @@ class Metric:
                     return int(self.period[:-len(unit)].strip()) * mul
         return 0
 
+    def extract_value_from_field(self, field: str) -> float:
+        """
+        Extract numeric value from field, handling key-value pairs and plain values
+
+        Returns 0.0 if the value cannot be converted to float.
+        """
+        # If it's a key-value pair, extract the value
+        if '=' in field:
+            key, value = field.split('=', 1)
+            # Remove quotes if present
+            value = value.strip('"')
+            try:
+                return float(value)
+            except ValueError:
+                return 0.0
+        else:
+            # Plain value
+            try:
+                return float(field)
+            except ValueError:
+                return 0.0
+
     def estimate_columns_value(self, fields):
-        '''
+        """
         Apply some arithmetic on several columns if needed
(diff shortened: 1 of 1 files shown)


igcollect/logfile_values.py
```diff
@@ -65,28 +65,66 @@ def get_timeshift(self):
                     return int(self.period[:-len(unit)].strip()) * mul
         return 0
 
+    def extract_value_from_field(self, field):
+        '''
+        Extract numeric value from field, handling key-value pairs and plain values
+        '''
```
kofrezo, line 71: It is fine to use single or double quotes in a project or script as long as it is consistent but doc strings should always be triple quotes. 

> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings.
https://peps.python.org/pep-0257/

igcollect/logfile_values.py
```diff
@@ -65,28 +65,66 @@ def get_timeshift(self):
                     return int(self.period[:-len(unit)].strip()) * mul
         return 0
 
+    def extract_value_from_field(self, field):
```
kofrezo, line 68:
```suggestion
    def extract_value_from_field(self, field: str) -> float:
```

igcollect/logfile_values.py
```diff
@@ -65,28 +65,66 @@ def get_timeshift(self):
...
+        '''
+        # If it's a key-value pair, extract the value
+        if '=' in field:
+            key, value = field.split('=', 1)
+            # Remove quotes if present
+            value = value.strip('"')
+            try:
+                return float(value)
+            except ValueError:
+                return 0
```
kofrezo, line 80:
```suggestion
                return 0.0
```

igcollect/logfile_values.py
```diff
@@ -65,28 +65,66 @@ def get_timeshift(self):
...
+        else:
+            # Plain value
+            try:
+                return float(field)
+            except ValueError:
+                return 0
```
kofrezo, line 86:
```suggestion
                return 0.0
```

igcollect/logfile_values.py
```diff
@@ -65,28 +65,66 @@ def get_timeshift(self):
                     return int(self.period[:-len(unit)].strip()) * mul
         return 0
 
+    def extract_value_from_field(self, field):
+        '''
+        Extract numeric value from field, handling key-value pairs and plain values
```
kofrezo, line 70:
```suggestion
        Extract numeric value from field, handling key-value pairs and plain values
        
        Returns 0.0 if the value cannot be converted to float.
```

igcollect/logfile_values.py
```diff
@@ -65,28 +65,66 @@ def get_timeshift(self):
...
+            try:
+                return float(field)
+            except ValueError:
+                return 0
+
     def estimate_columns_value(self, fields):
         '''
         Apply some arithmetic on several columns if needed
-        Warning: Estimates regardless of arithmetics rules
+        Now handles key-value pairs properly
```
kofrezo, line 91: That should rather go to the commit or changelog.

igcollect/logfile_values.py
```diff
@@ -65,28 +65,66 @@ def get_timeshift(self):
...
                 elif value == '+':
-                    result = result + float(fields[int(arr[index + 1])])
+                    field = fields[int(arr[index + 1])]
+                    result = result + self.extract_value_from_field(field)
                 elif value == '-':
-                    result = result - float(fields[int(arr[index + 1])])
-            except ZeroDivisionError:
+                    field = fields[int(arr[index + 1])]
+                    result = result - self.extract_value_from_field(field)
+            except (ZeroDivisionError, IndexError, ValueError):
```
kofrezo, line 127: You are checking whatever the divisor is zero so you don't need to except ZeroDivisionError anymore.

igcollect/logfile_values.py
```diff
@@ -170,12 +208,26 @@ def parse_args():
     return parser.parse_args()
 
 
+def extract_timestamp_value(field):
```
kofrezo, line 211: This isn't really exclusive to timestamps is it ? This could be any string value enclosed in quotes.

kofrezo (changes requested):
I would ask you to at least ensure we are consistently return floats and not a mix of integers and floats.

kofrezo (approved):

PR #202, opened 2026-03-19
kea: add Kea DHCP statistics collector

lamaral:


igcollect/kea.py
```diff
@@ -0,0 +1,72 @@
...
+        value = samples[0][0]
+        print(f'{args.prefix}.{sanitize_metric(name)} {value} {timestamp}')
+
+
+def get_stats(host: str, port: int) -> dict:
+    payload = json.dumps({
+        'command': 'statistic-get-all',
+    }).encode()
+    req = Request(
+        f'http://{host}:{port}/',
```
TuxPowered42, line 43: If an IPv6 address is given, do we need to wrap it it in `[]`?

igcollect/kea.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def sanitize_metric(name: str) -> str:
+    """Convert a Kea stat name to a Graphite-compatible metric path.
+
+    Examples:
+        pkt4-received               -> pkt4_received
+        subnet[1].assigned-addresses -> subnet.1.assigned_addresses
+        subnet[1].pool[0].total-addresses -> subnet.1.pool.0.total_addresses
+        subnet[1].pd-pool[0].total-pds    -> subnet.1.pd_pool.0.total_pds
```
TuxPowered42, line 69: Subnet IDs match Serveradmin IDs, correct?

lamaral: It's not enforced to be, but the use-cases I know were done as such.

igcollect/kea.py
lamaral: Python's urllib can handle IPv6 without needing the square brackets, but RFC 3986 specifies that they should be enclosed. 
I adjusted the code accordingly.

TuxPowered42 (approved):

PR #205, opened 2026-04-14
linux_nvme: add NVMe SMART collector

lamaral:
Discovers NVMe controllers via /sys/class/nvme and queries nvme smart-log for SMART metrics, emitting them in Graphite plaintext format. Model and serial names are sanitized (spaces and dots replaced with underscores) to produce valid metric paths. sysfs read failures are reported to stderr and skip the affected device without aborting the collection run.


raimund-s (approved):
looks good

PR #208, opened 2026-06-03
Fix LB Pools traffic collector

TuxPowered42:


igcollect/pf_labels.py
```diff
@@ -59,12 +59,25 @@ def parse_pf_labels(labels):
...
+        offset = len(labels)
+
+        # Only parse lines for lb_pools
+        if offset == 0:
+            continue
 
-        # Cut unnecessary things out of label
-        label = line_tab[0].split(":")[0]
-        label_re = POOL_RE.match(label)
+        for label in line_tab:
```
lamaral, line 68: You can iterate on `labels` from line 68 instead of `line_tab`.

igcollect/pf_labels.py
```diff
@@ -59,12 +59,25 @@ def parse_pf_labels(labels):
...
+            # Find the per-address family label
+            label_re_tmp = POOL_RE.match(label)
+            if label_re_tmp:
+                label_re = label_re_tmp
+                break
```
lamaral: I think you don't need the _tmp here. `re.match()` will return None if it doesn't match.

You can probably:
```python
if label_re := POOL_RE.match(label):
    break
```

Might be worthy also adding a `label_re = None` before the loop, because if no labels match, you will get label_re undefined on line 82.

lamaral: I'll take it back. label_re will get defined by walrus operator even if no matches at all are found, so no need to pre-define label_re before the loop.

igcollect/pf_labels.py
TuxPowered42: I don't think so. We need to count *all* labels first, and then figure out which one contains the address family. But maybe we can do it easier. We know the amount of data columns, so we don't need to know the amount of labels.

igcollect/pf_labels.py
TuxPowered42: removed the tmp variable

igcollect/pf_labels.py
TuxPowered42: Now the columns are checked from the end and labels from the start of the split line.

lamaral (approved):

